PR #52, opened 2018-06-15
Feature/no data filter

jtnachbar:


core/src/main/scala/astraea/spark/rasterframes/ml/NoDataFilter.scala
```diff
@@ -36,7 +38,12 @@ class NoDataFilter (override val uid: String) extends Transformer
   with HasInputCols with DefaultParamsWritable {
 
   def this() = this(Identifiable.randomUID("nodata-filter"))
-  final def setInputCols(value: Array[String]) = set(inputCols, value)
+  final def setInputCols(value: Array[String]): NoDataFilter = set(inputCols, value)
+  final def setInputCols(values: ArrayList[String]): NoDataFilter = {
+    val valueArr = Array[String](values:_*)
```
metasim: This is probably not going to make sense, but it's convention in the rest of the code, so plese change

```scala
setInputCols(value: Array[String]): NoDataFilter = ...
```
to:
```scala
setInputCols(value: Array[String]): this.type = ...
```

metasim: Also, I think you could change:

```scala
val valueArr = Array[String](values:_*)
```
to:
```scala
val valueArr = values.toArray
```

metasim:
Incorporated in bcf7745d23044e9a4f61dfe4540a5e0116ceb425

PR #54, opened 2018-06-20
Updated the GeoTrellis version to 2.0.0-RC1

jbouffard:
This PR updates the GeoTrellis version from `1.2.1` to `2.0.0-RC1` which is the most current version. In addition to the version update, minor changes were made to update the code.


jbouffard:
@metasim I'm not sure. I think they *should* work, but I'll need to test it and see.

datasource/src/main/scala/astraea/spark/rasterframes/datasource/geotiff/GeoTiffInfoSupport.scala
```diff
@@ -46,7 +46,7 @@ trait GeoTiffInfoSupport {
   }
 
   def extractGeoTiffLayout(reader: ByteReader): (GeoTiffReader.GeoTiffInfo, TileLayerMetadata[SpatialKey]) = {
-    val info = GeoTiffReader.readGeoTiffInfo(reader, false, true)
+    val info = GeoTiffReader.readGeoTiffInfo(reader, false, true, None)
```
pomadchin: Well at least because of this line its not compatible with 1.2

jbouffard: Oh, that's right. Thanks for pointing that out. @metasim Looks like these changes aren't compatible then.

pomadchin: Yea, that's because there were not `overviews` support in 1.2

metasim: We're not ready to upgrade our stack to GT 2.x, so I may need to release separate versions, or figure out a reflections-based approach to handling both cases.

Next week I'll merge this PR and then work on that so we can run against either version.

@pomadchin Is there any chance in 2.x to provide backwards compatibility, or have the semantics changed greatly?

pomadchin: If it would help you / us / @jbouffard  we can add an extra overload.

jbouffard: Yeah, that's something we could do if it's needed.

metasim:
I'm going to go ahead and merge this, and then on develop switch the default version back to GT 1.2.1 and Spark 2.2.1, and make sure the tests all pass for those versions. Then we can do a new PR for fixes relating to getting tests to pass for GT 2.0.x and Spark 2.3.1.

PR #58, opened 2018-06-25
Add maskByValue function to scala, sql, and py apis.

vpipkt:
Function to mask tile `T1` where tile `T2` equals `v`

Use cases would be cloud or snow masks from data producers. 


vpipkt:
I did not test or try out the Python API for `maskByValue`.

pyrasterframes/python/pyrasterframes/rasterfunctions.py
```diff
@@ -176,6 +176,7 @@ def _(geometryCol, srcCRSName, dstCRSName):
     'localAggDataCells': 'Compute the cellwise/local count of non-NoData cells for all Tiles in a column.',
     'localAggNoDataCells': 'Compute the cellwise/local count of NoData cells for all Tiles in a column.',
     'mask': 'Where the mask (second) tile contains NODATA, replace values in the source (first) tile with NODATA.',
+    'maskByValue': 'Where the mask (second) tile equals maskValue, replace values in the source (first) tile with NODATA',
```
metasim: @vpipkt This won't work right here (only works for `(Tile, Tile) => ?` methods). Can you just remove it, and we'll create a ticket for @bguseman to implement it in Python?

Issue #68, opened 2018-07-09
`spark.write.geotiff` round trip


vpipkt:
I am looking into using the geotiff writer and I discovered the following. A test resource geotiff (COG) is read and then without modification written in the `GeoTiffDataSourceSpec`.  The  output of the writer is distorted... I'll provide links to previews of the images as well as the output that I get for reference.

The output tif is larger, as entire tiles are written. It has the same datatype and nodata value for each band.

The gdalinfo output for both is below. The writer outputs a file named `example-geotiff.tiff`

```
$ gdalinfo core/src/test/resources/LC08_RGB_Norfolk_COG.tiff
Driver: GTiff/GeoTIFF
Files: core/src/test/resources/LC08_RGB_Norfolk_COG.tiff
Size is 1028, 989
Coordinate System is:
PROJCS["WGS 84 / UTM zone 18N",
    GEOGCS["WGS 84",
        DATUM["WGS_1984",
            SPHEROID["WGS 84",6378137,298.257223563,
                AUTHORITY["EPSG","7030"]],
            AUTHORITY["EPSG","6326"]],
        PRIMEM["Greenwich",0,
            AUTHORITY["EPSG","8901"]],
        UNIT["degree",0.0174532925199433,
            AUTHORITY["EPSG","9122"]],
        AUTHORITY["EPSG","4326"]],
    PROJECTION["Transverse_Mercator"],
    PARAMETER["latitude_of_origin",0],
    PARAMETER["central_meridian",-75],
    PARAMETER["scale_factor",0.9996],
    PARAMETER["false_easting",500000],
    PARAMETER["false_northing",0],
    UNIT["metre",1,
        AUTHORITY["EPSG","9001"]],
    AXIS["Easting",EAST],
    AXIS["Northing",NORTH],
    AUTHORITY["EPSG","32618"]]
Origin = (364455.000000000000000,4109985.000000000000000)
Pixel Size = (30.000000000000000,-30.000000000000000)
Metadata:
  AREA_OR_POINT=Area
Image Structure Metadata:
  COMPRESSION=DEFLATE
  INTERLEAVE=PIXEL
Corner Coordinates:
Upper Left  (  364455.000, 4109985.000) ( 76d31'33.12"W, 37d 7'35.07"N)
Lower Left  (  364455.000, 4080315.000) ( 76d31'13.92"W, 36d51'32.56"N)
[...]

PR #70, opened 2018-07-12
Fix for 68

metasim:


vpipkt:
I'm looking at this now... more to come

vpipkt:
Some inconsistencies seem to remain. 

Now if you observe the last Row of tiles in the output (those whose tiles are 221 rows tall), you will see some stretching of the features there, as if the tile was 256 cells tall.

metasim:
@vpipkt Thanks for the juxtapose view.... I never would have picked up on that.

metasim:
I'm going to merge this, but keep the bug open, so as to refine from here.

vpipkt:
Looks like #68 got closed?

PR #86, opened 2018-12-04
reflectively call readGeoTiffInfo with the correct args in geotrells 2.0

philvarner:


metasim:
Thank you *so* much!!

Issue #98, opened 2019-02-14
Snake case for `cellTypes` to `cell_types`


vpipkt:
In the Python API: https://github.com/locationtech/rasterframes/blob/develop/pyrasterframes/python/pyrasterframes/rasterfunctions.py#L157

PR #103, opened 2019-02-25
Add function references for SQL and Python API's.

vpipkt:


vpipkt:
Now need to update to `cell_types` from `cellTypes` since #98 fixed by #100

PR #111, opened 2019-04-23
Column functions to check for_all cells and exists cells

vpipkt:
Naming of functions TBD. Considerations:

1) `forall` and `exists` are pretty common key words in scala. For scala API users if we use those terms we want the mental model of the function to match to what's happening
2) `all` is a keyword in 
3) in numpy there are methods on arrays: `a.all()` and `a.any()` that do similar to the intent here
4) Sometimes we prefix a column by `tile_` if the name would otherwise be pretty common (`tile_sum`, `tile_max`).

Another thought: `is_no_data_tile` does something like `for_all` proposed here, testing the cells are all nodata. If we also had a [`undefined`](https://geotrellis.github.io/scaladocs/latest/index.html#geotrellis.raster.mapalgebra.local.Undefined$@apply(r:geotrellis.raster.Tile):geotrellis.raster.Tile); `is_no_data_tile(t)` is equivalent to `for_all(undefined(t))`

https://s22s.myjetbrains.com/youtrack/issue/RF-18


vpipkt:
Issue about defined / undefined : https://s22s.myjetbrains.com/youtrack/issue/AEP-328

PR #114, opened 2019-04-30
No Forking in Tests

jamesmcclain:


metasim:
@jamesmcclain Why did you need to turn forking off? My experience is that it causes more problems than it solves.

jamesmcclain:
> @jamesmcclain Why did you need to turn forking off? My experience is that it causes more problems than it solves.

No, I was just testing to see if that would help.  It has been known to fix issues like the one that we have been discussing.

PR #119, opened 2019-05-24
Ability to spatially join two RasterFrames with disparate CRS sets

metasim:


core/src/main/scala/org/locationtech/rasterframes/expressions/aggregates/TileRasterizerAggregate.scala
```diff
@@ -0,0 +1,95 @@
...
+import geotrellis.raster.{ArrayTile, CellType, MutableArrayTile, Raster, Tile}
+import geotrellis.vector.Extent
+import org.apache.spark.sql.expressions.{MutableAggregationBuffer, UserDefinedAggregateFunction}
+import org.apache.spark.sql.types.{DataType, StructField, StructType}
+import org.apache.spark.sql.{Column, Row, TypedColumn}
+import geotrellis.raster.reproject.Reproject
+import org.locationtech.rasterframes.expressions.aggregates.TileRasterizerAggregate.ProjectedRasterDefinition
+
+/**
+  * Aggregation function for creating a single raster from tiles.
```
vpipkt: May I suggest styling here a "a single Raster" possibly with ref to the gt class

core/src/test/scala/org/locationtech/rasterframes/RasterJoinSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+import org.locationtech.rasterframes.encoders.serialized_literal
+import org.locationtech.rasterframes.expressions.aggregates.TileRasterizerAggregate
+import org.locationtech.rasterframes.expressions.aggregates.TileRasterizerAggregate.ProjectedRasterDefinition
+
+
+class RasterJoinSpec extends TestEnvironment with TestData with RasterMatchers {
+  import spark.implicits._
+  describe("Raster join between two DataFrames") {
+    val s1 = readSingleband("L8-B4-Elkton-VA.tiff")
+    val s2 = readSingleband("L8-B4-Elkton-VA-4326.tiff")
```
vpipkt: This tiff has no nodata defined on it,  and in debugger I see cellType is `uint16raw` but there are nodata artifacts from the different griddings. See the black areas with values of 0 in the attached. 

The minimum value in the original file is about 6000.

image

vpipkt: I think the 0 / nodata issue here does not actually affect any of the tests currently written.

core/src/test/scala/org/locationtech/rasterframes/RasterJoinSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+    val r2 = s2.projectedRaster.mapTile(_.convert(IntConstantNoDataCellType)).toRF(10, 10)
+      .withExtent()
+      .withColumn("crs", serialized_literal(s2.crs))
+      .withColumnRenamed("tile", "tile2")
+
+    it("should join the same scene correctly") {
+      val joined = r1.rasterJoin(r1.withColumnRenamed("tile", "tile2"))
+      joined.count() should be (joined.count())
+
+      val measure = joined.select(rf_tile_mean(rf_local_subtract($"tile", $"tile2")))
```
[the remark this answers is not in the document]
vpipkt: I tried this a little and what I see is the `$"tile2"` contains the same cell values in all rows. Not sure why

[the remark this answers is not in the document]
metasim: 1) Yes, we need a test here. Was struggling to find something that works (hence our discussion on Friday).
2) Very possibly a bug.... however, how were you measuring that?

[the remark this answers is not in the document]
vpipkt: 1. It's my reading of this test case that we are joining the data to itself, so a trivial case where we should expect the same result back? 

2. I printed out the rf_tile_means and get the same tile mean for all rows. Not quite believing I printed out the rf_render_matrix...

core/src/test/scala/org/locationtech/rasterframes/RasterJoinSpec.scala
metasim: Still, I wish L8 had a NoData value set it in it... PITA.

core/src/main/scala/org/locationtech/rasterframes/extensions/RasterJoin.scala
```diff
@@ -0,0 +1,103 @@
...
+      d.merge(leftExtent, t.extent, t.tile, projOpts.method)
+    )
+  }).withName("reproject_and_merge")
+
+  def apply(left: DataFrame, right: DataFrame): DataFrame = {
+    val df = apply(left, right, left("extent"), left("crs"), right("extent"), right("crs"))
+    df.drop(right("extent")).drop(right("crs"))
+  }
+
+  def apply(left: DataFrame, right: DataFrame, leftExtent: Column, leftCRS: Column, rightExtent: Column, rightCRS: Column): DataFrame = {
```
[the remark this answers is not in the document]
metasim: kill joy? 😆

core/src/test/scala/org/locationtech/rasterframes/RasterJoinSpec.scala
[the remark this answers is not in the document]
metasim: 👍 
I should have indicated that the current test cases are wholly inadequate....

metasim:
@vpipkt thanks *so* much for extending the tests. Very much appreciated.

vpipkt:
Late arriving comment, could we add `rasterJoin` to pyrasterframes.RasterFrame ? I'm willing to take it on

core/src/main/scala/org/locationtech/rasterframes/extensions/RasterJoin.scala
```diff
@@ -0,0 +1,124 @@
...
+      col(leftExtent2), col(leftCRS2), col(t.columnName), col(rightExtent2), col(rightCRS2), rf_dimensions(leftTile)
+    ) as t.columnName)
+    val finalCols = leftAggCols.map(c => col(c.columnName)) ++ reprojCols
+
+    // Here's the meat:
+    left
+      // 1. Add a unique ID to each LHS row for subequent grouping.
+      .withColumn(id, monotonically_increasing_id())
+      // 2. Perform the left-outer join
+      .join(right, joinExprs, joinType = "left")
```
vpipkt: I believe this requires spark conf `spark.sql.crossJoin.enabled=true` .... should we try to warn the user ahead of time, or should we enable that when doing the withRasterFrames spark session goodies?

metasim: My understanding was that this should only be required if we were joining a DF on itself or doing an outer join with no predicate.... I had enabled it in the `TestEnvironment`, but we know how that worked out 🙄

core/src/main/scala/org/locationtech/rasterframes/extensions/DataFrameMethods.scala
```diff
@@ -172,25 +172,53 @@ trait DataFrameMethods[DF <: DataFrame] extends MethodExtensions[DF] with Metada
...
+    * @param right right dataframe
     * @param leftExtent this (left) dataframe's Extent column
     * @param leftCRS this (left) datafrasme's CRS column
-    * @param right right dataframe
     * @param rightExtent right dataframe's CRS extent
     * @param rightCRS right dataframe's CRS column
     * @return joined dataframe
     */
-  def rasterJoin(leftExtent: Column, leftCRS: Column, right: DataFrame, rightExtent: Column, rightCRS: Column): DataFrame =
+  def rasterJoin(right: DataFrame, leftExtent: Column, leftCRS: Column, rightExtent: Column, rightCRS: Column): DataFrame =
```
vpipkt: 👍

vpipkt:
@metasim I think we are all done once CI build passes.

PR #120, opened 2019-05-29
 Added rf_inverse_mask_by_value

metasim:


core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/Mask.scala
```diff
@@ -143,4 +143,24 @@ object Mask {
     def apply(srcTile: Column, maskingTile: Column, maskValue: Column): TypedColumn[Any, Tile] =
       new Column(MaskByValue(srcTile.expr, maskingTile.expr, maskValue.expr)).as[Tile]
   }
+
+  @ExpressionDescription(
+    usage = "_FUNC_(target, mask, maskValue) - Generate a tile with the values from the data tile, but where cells in the masking tile DO NOT contain the masking value, replace the data value with NODATA.",
+    arguments = """
+  Arguments:
+    * target - tile to mask
+    * mask - masking definition""",
```
vpipkt: missing the `maskValue` arg
```suggestion
    * mask - masking definition
    * maskValue - value in the `mask` for which to mark `target` as data cells
    """,
```

Issue #131, opened 2019-06-03
Miscellaneous/random `BufferUnderflowException`


metasim:
Context:

* It takes about an hour or so into an analysis to happen
* It doesn't happen all the time though
* Probably about half the time I get this exception

```
Caused by: java.nio.BufferUnderflowException
    at java.nio.HeapByteBuffer.get(HeapByteBuffer.java:151)
    at java.nio.ByteBuffer.get(ByteBuffer.java:715)
    at geotrellis.util.StreamingByteReader.getBytes(StreamingByteReader.scala:120)
    at geotrellis.raster.io.geotiff.LazySegmentBytes.getBytes(LazySegmentBytes.scala:120)
    at geotrellis.raster.io.geotiff.LazySegmentBytes$$anonfun$readChunk$1.apply(LazySegmentBytes.scala:99)
    at geotrellis.raster.io.geotiff.LazySegmentBytes$$anonfun$readChunk$1.apply(LazySegmentBytes.scala:97)
    at scala.collection.immutable.List.map(List.scala:273)
    at geotrellis.raster.io.geotiff.LazySegmentBytes.readChunk(LazySegmentBytes.scala:97)
    at geotrellis.raster.io.geotiff.LazySegmentBytes$$anonfun$getSegments$1.apply(LazySegmentBytes.scala:115)
    at geotrellis.raster.io.geotiff.LazySegmentBytes$$anonfun$getSegments$1.apply(LazySegmentBytes.scala:115)
    at scala.collection.Iterator$$anon$12.nextCur(Iterator.scala:434)
    at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:440)
    at scala.collection.Iterator$$anon$13.hasNext(Iterator.scala:461)
    at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
    at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:439)
    at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
    at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
    at geotrellis.contrib.vlm.geotiff.GeoTiffRasterSource.read(GeoTiffRasterSource.scala:58)
    at org.locationtech.rasterframes.ref.RasterSource$DelegatingRasterSource.read(RasterSource.scala:237)
    at org.locationtech.rasterframes.ref.RasterRef.realizedTile$lzycompute(RasterRef.scala:56)
[...]

PR #132, opened 2019-06-03
Experiment with read retries to handle BufferUnderflowExceptions.

metasim:
To address #131

### DIFF
--- a/core/src/main/scala/org/locationtech/rasterframes/ref/RasterSource.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/ref/RasterSource.scala
@@ -191,7 +191,25 @@ object RasterSource extends LazyLogging {
   /** A RasterFrames RasterSource which delegates most operations to a geotrellis-contrib RasterSource */
   abstract class DelegatingRasterSource(source: URI, delegateBuilder: () => GTRasterSource) extends RasterSource with URIRasterSource {
     @transient
-    lazy val delegate = delegateBuilder()
+    @volatile
+    private var _delRef: GTRasterSource = _
+
+    private def retryableRead[R >: Null](f: GTRasterSource => R): R = synchronized {
+      try {
+        if (_delRef == null)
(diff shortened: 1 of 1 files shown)


courtney-whalen:
[summer_crop_model (2).json.zip](https://github.com/locationtech/rasterframes/files/3249069/summer_crop_model.2.json.zip)

metasim:
According to @courtney-whalen large model training completed after this change.

courtney-whalen:
Yeah, tested it out several times and haven't gotten a BufferUnderflowException again

core/src/main/scala/org/locationtech/rasterframes/ref/RasterSource.scala
```diff
@@ -221,21 +239,49 @@ object RasterSource extends LazyLogging {
...
+//              throw be
+//            }
+//            else Thread.sleep(100)
+//            ()
+//        }
+//      }
+//      result
+//    }
+
+
```
metasim: Sorry about that comment block. Will get rid of eventually.

core/src/main/scala/org/locationtech/rasterframes/ref/RasterSource.scala
```diff
@@ -191,7 +191,25 @@ object RasterSource extends LazyLogging {
...
+      catch {
+        // On this exeception we attempt to recreate the delegate and read again.
+        case be: java.nio.BufferUnderflowException =>
+          _delRef = null
+          val newDel = delegateBuilder()
+          val result = f(newDel)
+          _delRef = newDel
+          result
+      }
+    }
```
metasim: This is the tricky bit. Need to make sure it's thread safe, even though multiple threads "shouldn't" be calling the read code.

pomadchin:
@metasim did it finally work?

metasim:
@pomadchin Not sure yet.... takes a larger deployment process and several hours of training to be absolutely sure.

PR #133, opened 2019-06-04
Additional Python source reorganization

metasim:
Sets the stage for python-based documentation build.


vpipkt:
@metasim 

One point of order. 

Travis build passes but it never actually runs the Python tests. It dies on trying to import pathlib. which I suspect means it is actually running python 2.7, but not sure.

So two issues there 

1. make the python tests correctly return as failed. 
2. more travis tweaking to run on python 3 if that's what we want.

https://travis-ci.org/locationtech/rasterframes/builds/541817162#L4430-L4434

Issue #140, opened 2019-06-07
Python UDT.test_udf_np_implicit_type_conversion fails


metasim:
As of 0acadda44c9ee75763ffbeb341ed258af0bafa48

```
=================================== FAILURES ===================================
___________________ UDT.test_udf_np_implicit_type_conversion ___________________

self = <tests.PyRasterFramesTests.UDT testMethod=test_udf_np_implicit_type_conversion>

    def test_udf_np_implicit_type_conversion(self):
        import math
        import pandas

        a1 = np.array([[1, 2], [0, 4]])
        t1 = Tile(a1, CellType.uint8())
        exp_array = a1.astype('>f8')

        @udf(TileUDT())
        def times_pi(t):
            return t * math.pi

        @udf(TileUDT())
        def divide_pi(t):
            return t / math.pi

        @udf(TileUDT())
        def plus_pi(t):
            return t + math.pi

        @udf(TileUDT())
        def less_pi(t):
            return t - math.pi

        df = self.spark.createDataFrame(pandas.DataFrame([{"tile": t1}]))
        r1 = df.select(
>           less_pi(divide_pi(times_pi(plus_pi(df.tile))))
        ).first()[0]

tests/PyRasterFramesTests.py:446:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.eggs/pyspark-2.3.3-py2.7.egg/pyspark/sql/dataframe.py:1144: in first
    return self.head()
.eggs/pyspark-2.3.3-py2.7.egg/pyspark/sql/dataframe.py:1132: in head
    rs = self.head(1)
.eggs/pyspark-2.3.3-py2.7.egg/pyspark/sql/dataframe.py:1134: in head
    return self.take(n)
.eggs/pyspark-2.3.3-py2.7.egg/pyspark/sql/dataframe.py:504: in take
    return self.limit(num).collect()
.eggs/pyspark-2.3.3-py2.7.egg/pyspark/sql/dataframe.py:466: in collect
    sock_info = self._jdf.collectToPython()
/anaconda3/envs/py27/lib/python2.7/site-packages/py4j/java_gateway.py:1257: in __call__
    answer, self.gateway_client, self.target_id, self.name)
.eggs/pyspark-2.3.3-py2.7.egg/pyspark/sql/utils.py:63: in deco
    return f(*a, **kw)
[...]

PR #141, opened 2019-06-07
Fixes #136 and #140.

### LINKED ISSUE #136: Tests fail under Python 2 but are not detected as such.
@metasim 

One point of order. 

Travis build passes but it never actually runs the Python tests. It dies on trying to import pathlib. which I suspect means it is actually running python 2.7, but not sure.

So two issues there 

1. make the python tests correctly return as failed. 
2. more travis tweaking to run on python 3 if that's what we want.

https://travis-ci.org/locationtech/rasterframes/builds/541817162#L4430-L4434

_Originally posted by @vpipkt in https://github.com/locationtech/rasterframes/pull/133#issuecomment-499501864_

metasim:


.circleci/config.yml
```diff
@@ -148,11 +148,11 @@ workflows:
...
-      - publish:
-          filters:
-            branches:
-              only:
-                - develop
+#      - publish:
+#          filters:
+#            branches:
+#              only:
+#                - develop
```
metasim: Publishing fails due to lacking LocationTech credentials. Don't want to be publishing SNAPSHOTS there anyway.

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
```diff
@@ -345,6 +345,8 @@ def __truediv__(self, right):
             ct = ct.with_no_data_value(quot.fill_value)
         return Tile(quot, ct)
 
+    def __div__(self, right):
+        return self.__truediv__(right)
```
metasim: Fix for #140. Python 2.7 compatibility.

PR #146, opened 2019-06-12
Python TileUDT and rf_type.Tile improvements 

vpipkt:


pyrasterframes/src/test/python/PyRasterFramesTests.py
```diff
@@ -403,6 +409,64 @@ def test_tile_udt_serialization(self):
             long_trip = df.first()["tile"]
             self.assertEqual(long_trip, a_tile)
 
+    def test_udf_on_tile_type_input(self):
+        import numpy.testing
+        # rf_local_add(t, 0) is to force lazy eval; accessing tile.tile is to get at the actual Tile type vs PRT struct
+        df = self.spark.read.rastersource(self.img_uri).withColumn('tile2', rf_local_add_int(col('tile.tile'), 0))
```
vpipkt: What can we do to reduce this kind of  friction?

pyrasterframes/src/test/python/PyRasterFramesTests.py
```diff
@@ -403,6 +409,64 @@ def test_tile_udt_serialization(self):
...
+        import numpy.testing
+
+        rf = self.rf
+
+        # create a trivial UDF that does something we already do with a raster_functions
+        @udf(TileUDT())
+        def my_udf(t):
+            import numpy as np
+            a = np.log1p(t.cells)
+            return Tile(a, CellType.from_numpy_dtype(a.dtype))
```
vpipkt: This kind of pattern is showing up more and more. See also our Tile.__add__, Tile.__sub__ etc.

vpipkt:
Marked WIP going to add a few more goodies.

pyrasterframes/src/test/python/PyRasterFramesTests.py
metasim: Yeh, that sucks..... Feel like we need to have a design discussion about it and all the datasource schemas, what the expectations are, is this a udf only problem, etc.

pyrasterframes/src/test/python/PyRasterFramesTests.py
```diff
@@ -403,6 +409,64 @@ def test_tile_udt_serialization(self):
...
+            (rf_dimensions('tile').cols.cast('int') * rf_dimensions('tile').rows.cast('int')).alias('expected'),
+            my_udf('tile').alias('result')).toPandas()
+
+        numpy.testing.assert_array_equal(
+            rf_result.expected.tolist(),
+            rf_result.result.tolist()
+        )
+
+        df_result = df.select(
+            (rf_dimensions(df.tile).cols.cast('int') * rf_dimensions(df.tile).rows.cast('int') -
```
metasim: Do we need an `rf_size` function?

pyrasterframes/src/test/python/PyRasterFramesTests.py
```diff
@@ -403,6 +409,64 @@ def test_tile_udt_serialization(self):
...
+    def test_udf_on_tile_type_output(self):
+        import numpy.testing
+
+        rf = self.rf
+
+        # create a trivial UDF that does something we already do with a raster_functions
+        @udf(TileUDT())
+        def my_udf(t):
+            import numpy as np
+            return Tile(np.log1p(t.cells))
```
metasim: win!

pyrasterframes/src/test/python/PyRasterFramesTests.py
metasim: If it's a `udf` only problem, this is where I wonder if a `rf_udf` annotation could be used/constructed/registered that handles the extraction.

Even better would be if we could handle this through automatic Catalyst casts/conversions over Python UDFs but needs more research.

pyrasterframes/src/test/python/PyRasterFramesTests.py
metasim: My hope is that we can make `udf`s a true path of last resort... what can we do to get closer to feature complete with columnar functions?

pyrasterframes/src/test/python/PyRasterFramesTests.py
metasim: Tracked via https://github.com/locationtech/rasterframes/issues/147

pyrasterframes/src/test/python/PyRasterFramesTests.py
vpipkt: Indeed. i think the familiar model here is that one *could* iterate over values in a numpy array but it is uncommon in practice because the built in functions are more convenient and known to be faster. I am glad the user gets some rope. But we also have to make this point clear in docs.

PR #148, opened 2019-06-14
Petty render python `Tile` in IPython / Jupyter

vpipkt:


pyrasterframes/src/main/python/pyrasterframes/rf_types.py
```diff
@@ -349,9 +349,102 @@ def __div__(self, right):
...
+        axis.imshow(self.cells)
+        axis.set_aspect('equal')
+        axis.xaxis.set_ticks([])
+        axis.yaxis.set_ticks([])
+
+        axis.set_title('{}, {})'.format(self.dimensions(), self.cell_type.__repr__()))  # compact metadata as title
+
+        with io.BytesIO() as output:
+            canvas.print_png(output)
+            return output.getvalue()
```
metasim: If we're truly rendering to PNG, it should happen in the JVM where it's a lot faster.

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
```diff
@@ -349,9 +349,102 @@ def __div__(self, right):
         return self.__truediv__(right)
 
     def dimensions(self):
-        # list of cols, rows as is conventional in GeoTrellis and RasterFrames
+        """ Return a list of cols, rows as is conventional in GeoTrellis and RasterFrames."""
         return [self.cells.shape[1], self.cells.shape[0]]
 
+    def to_png(self):
```
metasim: I feel like this deserves a "separation of concerns" analysis.

My preference would be to have a separate module that is optionally imported to either add `foo(Tile)` functions or inject extensions on `Tile` class, and keep this focused on the minimal representation of a `Tile` in Python.

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
metasim: Seems like there's a lot more going on than just rendering a PNG... what's with the `Figure` and `axis` settings?

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
```diff
@@ -349,9 +349,102 @@ def __div__(self, right):
...
+
+
+try:
+    from IPython import get_ipython
+    # modifications to currently running ipython session, if we are in one; these enable nicer visualization for Pandas
+    if get_ipython() is not None:
+        import pandas
+        html_formatter = get_ipython().display_formatter.formatters['text/html']
+        html_formatter.for_type(pandas.DataFrame, Tile.pandas_df_to_html)
+except ImportError:
```
metasim: Yeh, I definitely feel like this needs to be in a separate `rf_ipython` (or whatever) module.

pyrasterframes/src/main/python/requirements.txt
```diff
@@ -3,3 +3,4 @@ shapely
 pyspark==2.3.3
 numpy>=1.7
 pandas
+matplotlib<3.0.0  # no python 2.7 support after v2.x.x
```
metasim: What happens in our 2.7 environment? Does this run OK in CircleCI?

vpipkt: I think the build passes with this requirement.

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
vpipkt: Ahh i didn't realize we could do it on JVM side, but now i remember we used to do something like that... 

The Figure and axis are more about trying to get a certain output size and a particular styling with matplotlib, basically removing axis ticks and labels.

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
vpipkt: @metasim can you take another look. we retain the `_repr_png_` bound method, but that will (ostensibly) only be called by Ipython or jupyter;

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
vpipkt: The more I think about it the more I realize that it may not make sense to render PNG on the JVM... Because we only want to render PNG for the python `Tile` after we have serialized it from Catalyst.  Does that make sense?

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
```diff
@@ -344,10 +344,16 @@ def __matmul__(self, right):
         return Tile(np.matmul(self.cells, other))
 
     def dimensions(self):
-        # list of cols, rows as is conventional in GeoTrellis and RasterFrames
+        """ Return a list of cols, rows as is conventional in GeoTrellis and RasterFrames."""
         return [self.cells.shape[1], self.cells.shape[0]]
 
 
+    def _repr_png_(self):
```
metasim: I like this solution a lot more.

Did you want to reinstate `_repr_html_` since you have the code for it?

pyrasterframes/src/main/python/requirements.txt
metasim: Do we need ipython here too?

pyrasterframes/src/main/python/pyrasterframes/rf_ipython.py
```diff
@@ -0,0 +1,110 @@
...
+    axis.imshow(tile.cells)
+    axis.set_aspect('equal')
+    axis.xaxis.set_ticks([])
+    axis.yaxis.set_ticks([])
+
+    axis.set_title('{}, {})'.format(tile.dimensions(), tile.cell_type.__repr__()))  # compact metadata as title
+
+    with io.BytesIO() as output:
+        canvas.print_png(output)
+        return output.getvalue()
```
metasim: If we were still in DataFrame space, I'd push to do this work on the JVM side, adding a method to `PyRFContext` that essentially implements this:

https://github.com/locationtech/rasterframes/blob/136e222afbb3d1beff1d95e5d8587a23f5ff14b2/core/src/test/scala/examples/Masking.scala#L51-L54

(or [this](https://github.com/locationtech/rasterframes/blob/136e222afbb3d1beff1d95e5d8587a23f5ff14b2/docs/src/main/tut/exporting-rasterframes.md#exporting-a-raster))

The above just seems like an overly involved process to get a PNG representation when we've already got code to do it. However, I get that the `tile` parameter at this point has been already pulled out of the Spark context, and is fully represented in the Python VM, and pumping it back over the Py4J gateway is silly. So I guess I'll have to live with it (but not like it) ;-).

pyrasterframes/src/main/python/pyrasterframes/rf_ipython.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def tile_to_html(tile):
```
metasim: See comment below about `_repr_html_`.

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
metasim: Yes.

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
vpipkt: I chose not to because I am not sure how IPython decides which `_repr_*_` to use, and the html flavor is always going to be more expensive.

PR #151, opened 2019-06-18
Provide means of accommodating lazy tiles in Python

### LINKED ISSUE #147: Reduce / simplify ceremony to access `ProjectedRasterTile` in python
I [commented in a PR](https://github.com/locationtech/rasterframes/pull/146#pullrequestreview-248872236)  about python TileUDT on the following code block there.

```
# rf_local_add(t, 0) is to force lazy eval; accessing tile.tile is to get at the actual Tile type vs PRT struct
df = self.spark.read.rastersource(self.img_uri).withColumn('tile2', rf_local_add_int(col('tile.tile'), 0))
# downstream we collect df in the python driver
```

The problem is that on the python side this rastersource `tile` column is actually a struct, so the user has to overcome the friction of use `tile.tile` to get the `tile` type. 

The second challenge is the lazy evaluation. In the above snip if i did not add 0, I'd see the following error in  the `deserialize`. Obviously we need to do some null checking but better would be a mechanism for the python side to request the evaluation of the cell values.  Preferably this would happen without the user having to worry if the under the hood implementation is lazy.

```
/usr/local/rasterframes/pyrasterframes.zip/pyrasterframes/rf_types.py in deserialize(self, datum)
    416         cell_data_bytes = datum.cell_data.cells
    417         try:
--> 418             as_numpy = np.frombuffer(cell_data_bytes, dtype=cell_type.to_numpy_dtype())
    419             reshaped = as_numpy.reshape((rows, cols))
    420             t = Tile(reshaped, cell_type)

AttributeError: 'NoneType' object has no attribute '__buffer__'
```

metasim:
Closes #147


metasim:
@vpipkt Going ahead and merging this so it doesn't diverge too far. Feel free to comment and I'll roll changed into a separate PR.

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
```diff
@@ -414,6 +417,16 @@ def deserialize(self, datum):
         cols = datum.cell_context.dimensions.cols
         rows = datum.cell_context.dimensions.rows
         cell_data_bytes = datum.cell_data.cells
+        if cell_data_bytes is None:
+            if datum.cell_data.ref is None:
+                raise Exception("Invalid Tile structure. Missing cells and reference")
+            else:
+                payload = datum.cell_data.ref
+                cell_data_bytes = RFContext.active()._resolve_raster_ref(payload)
```
vpipkt: 👍  

And an extra 😹  for checking both ref and cells for null

PR #178, opened 2019-07-02
Upgraded gdal-warp-bindings

metasim:
Currently core dumps on MacOS.


metasim:
Depends on https://github.com/geotrellis/gdal-warp-bindings/issues/51

PR #205, opened 2019-07-24
Correct NoDataFilter implementation in Python; supervised ML doc example

vpipkt:
WIP because need to revisit all the verbiage. But I do have the supervised ML example running in about 165s on my laptop. Not too shabby.


pyrasterframes/src/main/python/docs/supervised-learning.pymd
```diff
@@ -0,0 +1,197 @@
...
+    rf_tile(df.B12).alias('B12'),
+)
+df.printSchema()
+```
+
+## Data Prep
+
+### Label Data
+
+[](https://github.com/locationtech/rasterframes/blob/develop/core/src/test/resources/L8-Labels-Elkton-VA.geojson)
```
metasim: I've been trying to use stable GitHub "raw" URLs, but it ends up being a PITA to get the real ones (due to the use of redirects). We should think of a better way post release.

vpipkt: here i was ok with the `blob` view because it does render nicely and it is merely for reader's convenience.

pyrasterframes/src/main/python/docs/supervised-learning.pymd
```diff
@@ -0,0 +1,205 @@
+# Supervised Machine Learning
+
+In this example we will demonstrate how to fit and score an unsupervised learning model with a sample of Sentinel-2 data and hand-drawn vector labels over different [land cover](https://en.wikipedia.org/wiki/Land_cover) types.
```
courtney-whalen: Should be "supervised" not "unsupervised"

metasim: @vpipkt Fix this before merge?

courtney-whalen:
@vpipkt I added some suggestions/comments - take it or leave it! I think the page looks good overall. Nice example!

pyrasterframes/src/main/python/docs/supervised-learning.pymd
vpipkt: Great catch .... 🤷‍♂

pyrasterframes/src/main/python/docs/supervised-learning.pymd
```diff
@@ -0,0 +1,205 @@
...
+from pyspark.ml.classification import DecisionTreeClassifier
+from pyspark.ml.evaluation import MulticlassClassificationEvaluator
+from pyspark.ml import Pipeline
+```
+
+SparkML requires that each observation be in its own row, and those observations be packed into a single [`Vector`](https://spark.apache.org/docs/latest/api/python/pyspark.ml.html#module-pyspark.ml.linalg) object. The first step is to "explode" the tiles into a single row per cell/pixel with the `TileExploder` (see also @ref:[`rf_explode_tiles`](reference.md#rf_explode_tiles)). Then we filter out any rows that have `NoData` values (which will cause an error during training). Finally we use the SparkML `VectorAssembler` to create that `Vector`. 
+
+It is worth discussing a couple of interesting things about the `NoDataFilter`. First we filter by the mask column. This achieves the filtering of observations only to the good pixels, without having to explicitly mask and assign NoData in all eleven columns of raster data. The other column specified is the `label` column. When it is time to score the model, the pipeline will ignore the fact that there is no `label` column on the input DataFrame.
```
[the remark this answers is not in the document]
vpipkt: ah i've left out the detail that the raster bands have no NoData.

pyrasterframes/src/main/python/docs/supervised-learning.pymd
```diff
@@ -0,0 +1,205 @@
...
+
+pipeline.getStages()
+```
+
+## Train the Model
+
+Push the "go button"! This will actually run each step of the Pipeline we created including fitting the decision tree model.
+
+```python, train
+model = pipeline.fit(df_mask.filter(rf_tile_sum('label') > 0).cache())
```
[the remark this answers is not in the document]
vpipkt: I don't think its actually logically necessary but maybe more efficient

pyrasterframes/src/main/python/docs/supervised-learning.pymd
```diff
@@ -0,0 +1,205 @@
...
+    rf_tile(df.B05).alias('B05'),
+    rf_tile(df.B06).alias('B06'),
+    rf_tile(df.B07).alias('B07'),
+    rf_tile(df.B08).alias('B08'),
+    rf_tile(df.B09).alias('B09'),
+    rf_tile(df.B11).alias('B11'),
+    rf_tile(df.B12).alias('B12'),
+)
+df.printSchema()
+```
```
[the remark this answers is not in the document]
vpipkt: Very good. We can try this when the GTiff writer is done. And I think a PNG writer is in the works too which probably works better for embedding in docs. I created an issue to improve docs in this way. See #217 

I'll try to do it with just the red green and blue Tiles as well.  That ought to work.

pyrasterframes/src/main/python/docs/supervised-learning.pymd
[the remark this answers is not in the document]
vpipkt: going to furhter refine this, also found a logic error in my code

vpipkt:
Made a variety of edits based on @courtney-whalen feedback and that's greatly appreciated.

I've not been able to fully evaluate the "knitted" document due to slow connection. But I'll take a look as soon as I can and merge...

pyrasterframes/src/main/python/docs/raster-write.pymd
```diff
@@ -33,7 +33,7 @@ display(tile) # IPython.display function
 
 Within an IPython or Jupyter interpreter a Pandas DataFrame containing a column of _tiles_ will be rendered as the samples discussed above. Simply import the `rf_ipython` submodule to enable enhanced HTML rendering of a Pandas DataFrame.
 
-In the example below, notice the result is limited to a small subset. For more discussion about why this is important, see the @ref:[Pandas and NumPy discussion](pandas-numpy.md).
```
vpipkt: this will fix  broken docs build

vpipkt:
The last part of the rendered previews will look about like this.

image
 
and then ... 

image

vpipkt:
@metasim can you take a peek. In the `makeSite` build I'm not seeing the ML page index.

So it's unclear to the reader that there is a supervised ML example unless they go all the way to the end of the unsupervised page. 

image

metasim:
This is what I ended up with.

Screen Shot 2019-07-29 at 12 58 08 PM

PR #212, opened 2019-07-26
Writing GeoTIFFs

metasim:


pyrasterframes/src/main/python/docs/raster-write.pymd
```diff
@@ -48,21 +50,47 @@ pandas_df.dtypes
...
-It is oftentimes convenient to have a simple overview image of the data, particularly if it spans a very wide area.
 
 ## GeoTIFFs
 
-GeoTIFF is one of the most common file formats for spatial data. 
+GeoTIFF is one of the most common file formats for spatial data, providing flexibility in data encoding, representation, and storage. RasterFrames provides a specialized Spark DataFrame writer for rendering a RasterFrame to a GeoTiff.
+
+One downside to GeoTIFF is that it is not a big data native format. To create a GeoTIFF all the data to be encoded has to be in the memory of one compute node (in Spark parlance, this is a "collect"), limiting it's maximum size substantially compared to that of a full cluster environment. When rendering GeoTIFFs in RasterFrames, you either need to specify the dimensions of the output raster, or be aware of how big the collected data will end up being.
+
+Fortunately, we can use the cluster computing capability to downlample the data into a more manageble saze. For sake of example, let's render a simple RGB overview image of our scene as a small raster:
```
vpipkt: spelling "saze".

pyrasterframes/src/main/python/docs/raster-write.pymd
```diff
@@ -48,21 +50,47 @@ pandas_df.dtypes
...
+
+```python write_geotiff
+import os.path
+from docs import docs_dir
+cat = """  
+red,green,blue
+{},{},{}
+""".format(scene('01'), scene('04'), scene('03'))
+outfile = os.path.join(docs_dir(), 'geotiff-overview.tif')
+rf = spark.read.raster(catalog=cat, catalog_col_names=['red', 'green', 'blue'])
```
vpipkt: would be nice to show how large the original raster is.

pyrasterframes/src/main/python/docs/raster-write.pymd
vpipkt: and can we mention the downsampling method? I assume it's nearest nieghbor? Is it configurable?

pyrasterframes/src/main/python/docs/raster-write.pymd
```diff
@@ -48,21 +50,47 @@ pandas_df.dtypes
...
+rf.write.geotiff(outfile, crs='EPSG:4326', raster_dimensions=(256, 256))
```
vpipkt: worth pointing out above that we are also reprojecting the raster here.

pyrasterframes/src/main/python/docs/raster-write.pymd
```diff
@@ -48,21 +50,47 @@ pandas_df.dtypes
...
+import rasterio
+import numpy as np
+from rasterio.plot import show, show_hist
+ 
+with rasterio.open(outfile) as src:
+    cells = np.clip(src.read(), 0, 1800).astype('float32')
+    show(cells)
+    show_hist(src, bins=50, lw=0.0, stacked=False, alpha=0.3,
+        histtype='stepfilled', title="Overview Histogram")
+```
```
vpipkt: I would also like to see an example where we select a contiguous subset and write it out at full resolution and in native CRS. 

Just for completeness. 

I think that would be a quite common use case also. Consider for example writing out ML classification model predictions for an area intersecting a polygon.

pyrasterframes/src/main/python/pyrasterframes/__init__.py
```diff
@@ -167,10 +173,32 @@ def temp_name():
...
+    df_writer,
+    path=None,
+    crs=None,
+    raster_dimensions=None,
+    **options):
+
+    if raster_dimensions is not None:
+        if isinstance(raster_dimensions, tuple):
+            options.update({
+                "imageWidth": raster_dimensions[0],
```
vpipkt: may make sense to check that the tuple length is 2 and that the members are `int` type?

pyrasterframes/src/main/python/pyrasterframes/__init__.py
```diff
@@ -167,10 +173,32 @@ def temp_name():
...
+                "imageHeight": raster_dimensions[1]
+            })
+        elif isinstance(raster_dimensions, str):
+            parts = raster_dimensions.split(',')
+            options.update({
+                "imageWidth": parts[0],
```
vpipkt: again maybe makes sense to check `len(parts)==2` and `all([isinstance(p, int) for p in parts])`

pyrasterframes/src/main/python/pyrasterframes/__init__.py
```diff
@@ -167,10 +173,32 @@ def temp_name():
...
+                "imageHeight": parts[1]
+            })
+
+    if crs is not None:
```
vpipkt: in our discussions before i thought crs was required?

vpipkt: Yea. 

```
Py4JJavaError: An error occurred while calling o107.save.
: java.lang.IllegalArgumentException: requirement failed: A destination CRS must be provided
	at scala.Predef$.require(Predef.scala:224)
```

So we could have a check like this in place of the if statement. 

```python
```suggestion
    assert(crs is not None, "You must pass in a `crs` argument")
```

 WOuld be good in docs to give some details about what's valid. EPSG codes? proj strings? wkt? What about pyproj [CRS object](https://pyproj4.github.io/pyproj/stable/api/crs.html)?  rasterio [CRS](https://rasterio.readthedocs.io/en/stable/api/rasterio.crs.html#rasterio.crs.CRS) Possibly some of those can come later in a future issue / PR.

vpipkt:
Excited that my initial test is looking very good!

vpipkt:
One interesting thing. 

I read and wrote to gtiff a single MODIS granule (single channel). In the resulting `gdalinfo` the pixel size is (460.720950959172399,-460.720950959172399) 

But the original granule's pixel size is slightly larger: (463.312716527916905,-463.312716527916677)

See the movie attached, you can kind of see an artifact that may be the tile boundaries.

Happy to share original and written file.


[ScreenRecordModis.mov.zip](https://github.com/locationtech/rasterframes/files/3448602/ScreenRecordModis.mov.zip)


The tile dimensions are pretty much normal for this kind of file, so pretty irregular. The most common one is not even square.

```
+------------------+-----+
|rf_dimensions(B01)|count|
+------------------+-----+
|         [256, 96]|    6|
|        [257, 257]|   18|
|         [257, 96]|    3|
|          [97, 96]|    1|
|         [97, 257]|    6|
|        [256, 257]|   36|
|        [257, 256]|    9|
|        [256, 256]|   18|
|         [97, 256]|    3|
+------------------+-----+
```

pyrasterframes/src/main/python/pyrasterframes/__init__.py
metasim: How do you do that? Nested `isinstance` calls?

metasim: Heh... nevermind.

pyrasterframes/src/main/python/pyrasterframes/__init__.py
metasim: > in our discussions before i thought crs was required?

The only case where it isn't is if it's already a "layer".... probably not the regular use case however.

pyrasterframes/src/main/python/docs/raster-write.pymd
metasim: Will add statement on method used. Making it configurable:
https://github.com/locationtech/rasterframes/issues/222

pyrasterframes/src/main/python/docs/raster-write.pymd
metasim: https://github.com/locationtech/rasterframes/issues/223

pyrasterframes/src/main/python/docs/raster-write.pymd
metasim: https://github.com/locationtech/rasterframes/issues/228

Issue #229, opened 2019-08-01
Unexpected behavior for rf_local_add with nodata tiles


emc5ud:
Hi, I was playing around with `rf_local_add` and came across the following results. I expected that with local tile arithmatic, no_data + data = no_data.

This code snippet creates a tile with 0's and 1's, and a tile with all 0's. Then I set 1 as the no data value. This makes the tile with 0's and 1's to now have 0's and "no datas" . Finally, I add the tiles together. The tile with with 0's and 1's does have no data as expected, but the sum does not have any nodata.
```
tile_size = 100
x = Tile(np.eye(tile_size, dtype='int16'))
y = Tile(np.zeros((tile_size, tile_size), dtype='int16'))

rf = spark.createDataFrame(pd.DataFrame(data = 
        {'eye_tile': [x],
        'zero_tile': [y]}))

mask_ct = CellType('uint16').with_no_data_value(1)
masked_rf = rf.withColumn('eye_tile_nd', 
                           rf_convert_cell_type('eye_tile', mask_ct.cell_type_name)) \
              .withColumn('zero_tile_nd',
                          rf_convert_cell_type('zero_tile', mask_ct.cell_type_name)) \
              .withColumn('masked', rf_local_add('eye_tile_nd', 'zero_tile_nd'))
```
![image](https://user-images.githubusercontent.com/8228717/62318434-da630900-b469-11e9-9a9e-dcdbcce549b5.png)

However, when doing the sum directly on the tiles, I do get a result with nodatas. 
```
r = masked_rf.collect()
r[0].eye_tile_nd + r[0].zero_tile_nd
```
![image](https://user-images.githubusercontent.com/8228717/62318518-01213f80-b46a-11e9-8d35-d264859de677.png)

The second behavior is what I expected, but regardless I think the behavior should be consistent.

vpipkt:
Indeed behavior in the `Tile` operators should be consistent with the column functions. 

@metasim what would you expect from GeoTrellis?

metasim:
GeoTrellis Does this:

* https://github.com/locationtech/geotrellis/blob/a463c48e7470e4ecfc0ae0ac297133d8bd00d641/raster/src/main/scala/geotrellis/raster/mapalgebra/local/Add.scala#L33

RasterFrames changes that behavior here:

[...]

PR #230, opened 2019-08-01
update nodata docs to show cell type conversion behavior and aggregations

emc5ud:


pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,92 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
 display(sample[1])
 ```
 
-## NoData in Arithmetic Operations
+## Changing a Tile's NoData Values
+
+One way to mask a tile is to make a new tile with a user defined NoData value. We will explore this method below. First, lets create a rasterframe from a tile with values of 0, 1, 2, and 3.
+
+```python create_dummy_tile
+tile_size = 100
```
metasim: Idea: If you make these smaller (say, 3 or 4) you could use the `rf_render_martix` function to show the contents of the tile within the dataframe.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,92 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+x = np.zeros((tile_size, tile_size), dtype='int16')
+
+for i in range(4):
+    x[:, i*tile_size//4:(i+1)*tile_size//4] = i
+x = Tile(x)
+
+rf = spark.createDataFrame(pd.DataFrame(data =
```
metasim: Is it necessary to go through Pandas to create a Spark dataframe? Would be nice to avoid teaching the user that ceremony if we can.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,137 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+```
+
+First, we mask the value of 1 by making a new column with the user defined cell type 'uint16ud1'. Then, we mask out the value of two by making a tile with the cell type 'uint16ud2'.
+
+```python
+def get_nodata_ct(nd_val):
+    return CellType('uint16').with_no_data_value(nd_val).cell_type_name
+
+masked_rf = rf.withColumn('tile_nd_1', 
+                           rf_convert_cell_type('tile', get_nodata_ct(1))) \
```
metasim: @vpipkt We should fix this to accept the `CellType` class directly, and not have the user do `cell_type_name`.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,137 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+## Changing a Tile's NoData Values
+
+One way to mask a tile is to make a new tile with a user defined NoData value. We will explore this method below. First, lets create a rasterframe from a tile with values of 0, 1, 2, and 3.
+
+```python create_dummy_tile
+tile_size = 100
+x = np.zeros((tile_size, tile_size), dtype='int16')
+
+for i in range(4):
+    x[:, i*tile_size//4:(i+1)*tile_size//4] = i
```
metasim: I have no idea what this is doing. @vpipkt Are we assuming that users are adept at NumPy? I'd prefer to have a comment that tells us the shape and content of what's being constructed here.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,137 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+collected = masked_rf.collect()
+```
 
-local algebra example; same celltype what happens to nodata 
-Possibly use st_geomFromWkt and rf_rasterize to create something to work from
+Let's look at the new Tiles we created. The tile named `tile_nd_1` has the 1 values masked out as expected. 
 
-agg
+```python
+display(collected[0].tile_nd_1)
```
metasim: @vpipkt Do we need to go through the docs and be consistent with usage of `display` vs. `show`?  I don't really understand why both exist, but it might be confusing to users not knowing which is the endorsed way... ?

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,137 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+```python
+masked_rf.select(rf_tile_sum('tile'), rf_tile_sum('tile_nd_1'), rf_tile_sum('tile_nd_2')).show()
+```
+
+
+## Combining Tiles with Different Data Types
+
+RasterFrames supports having Tile columns with multiple cell types in a single RasterFrame. It is important to understand how these different cell types interact. 
+
+Let's first create a RasterFrame that has columns of `float` and `int` cell type.
```
metasim: Isn't it `short` and `int`?

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: Yeah that was a little annoying

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: If you think thank is preferable I can do that. I personally like the rendered image.

emc5ud: image
You are suggesting something like this? I don't know how to deal with the ascii.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: Maybe Jason knows? From what I could tell using pandas was the cleanest.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: Maybe this is a little too fancy, but I mainly just wanted the user to ignore it since it and take the result for granted. I had hoped that the description above the cell block and the picture rendered below would show what that code section does. I could change the tile to have random values of 0, 1, 2, and 3 instead. Less pretty, but slightly simpler numpy.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: Yup that's an error on my part. Fixing.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: are you referring to spark's `show` method? I wasn't aware there was a `show` in Ipython similar to `display`

emc5ud: I've been using `display` whenever I want a tile rendered as an image

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: Nvm I didn't know you can set truncate to false. I personally still like the image but I'm open to whatever.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: After talking to Jason, he thinks that going through pandas is the clearest way of doing this. Most online tutorials with pyspark also involve making an intermediate pandas dataframe.

vpipkt: I think we could also do something like 

```
spark.createDataFrame([
   Row(tile=x),
])
```

I think also it is a matter of what the user is familiar with.

emc5ud: I think I like this approach actually. Changed it.

metasim: Yeh, that was the approach I prefer.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
metasim: @emc5ud I think a comment describing what you're creating is sufficient.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
metasim: Stick with the image... I wasn't paying attention and didn't think that `display` was rendering the tile, which i think is better in this case.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,134 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+              .withColumn('tile_nd_2',
+                          rf_convert_cell_type('tile_nd_1', get_nodata_ct(2))) \
+```
 
+```python
+collected = masked_rf.collect()
+```
 
-## Dealing with Multiple Cell Types
+Let's look at the new Tiles we created. The tile named `tile_nd_1` has the 1 values masked out as expected.
```
vpipkt: May be worth mentioning quickly here (and possibly above at the blue sentinel masking) that white values are NoData ?

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,134 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
-Quick demo of ND in two different cell types
+And the tile named `tile_nd_2` has the values of 1 and 2 masked out. This is because we created the tile by setting a new user defined NoData value to `tile_nd_1` the values previously masked out in `tile_nd_1` stayed masked when creating `tile_nd_2`.
+
+```python
+display(collected[0].tile_nd_2)
+```
+
+## Nodata Values in Aggregation
+
+Let's use the same tile as before to demonstrate how NoData values affect tile aggregations
```
vpipkt:
```suggestion
Let's use the same tile as before to demonstrate how NoData values affect tile aggregations.
```

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,134 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+
+```python
+tile_size = 100
+x = np.zeros((tile_size, tile_size), dtype='int16')
+for i in range(4):
```
vpipkt: we don't need to redifine `x` from the previous chunk. Also don't need to redefine `rf`

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,134 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+x = Tile(x)
+
+rf = spark.createDataFrame([Row(tile=x)])
+display(x)
+```
+
+First we create the two new masked tile columns as before. One with only the value of 1 masked, and the other with and values of 1 and 2 masked. 
+
+```python
+def get_nodata_ct(nd_val):
```
vpipkt: also don't need to redifeind this function from previous chuncks.

vpipkt: I'm actually now gettin a little lost with what you are intending to do differently 😜

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,134 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+                           rf_convert_cell_type('tile', get_nodata_ct(1))) \
+              .withColumn('tile_nd_2',
+                          rf_convert_cell_type('tile_nd_1', get_nodata_ct(2)))
+```
 
+The results of `rf_tile_sum` vary on the tiles that were masked. This is because any cells with NoData values are ignored in the aggregation. Note that `tile_nd_2` has the lowest sum, since it has the fewest amount of data cells.
 
+```python
+masked_rf.select(rf_tile_sum('tile'), rf_tile_sum('tile_nd_1'), rf_tile_sum('tile_nd_2')).show()
+```
```
vpipkt: It would be worthwhile to perhaps add a level 3 header `### ` differentiating the Tile Aggregate, DataFrame Aggregate, and Local Aggregate behaviors. Here you have done Tile Agg, could expand to include the other 2 or just point to the aggreagation.md docs page 

Another thought here it may be worth working through the artithmetic for one of the cases to show the reader how to work out the expected sum. it's a block of 25*100 * value for each, right.

FWIW also there is a new section we could add  about ND in local algebra; the bug https://github.com/locationtech/rasterframes/issues/229 is now fixed to rf_local_add(t1, t2) will work like d + nd = nd.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,134 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+```python
+rf_added = rf.withColumn('xy_sum', rf_local_add('y', 'x'))
+rf_added.select(rf_cell_type('xy_sum'), rf_cell_type('y'), rf_cell_type('x')).distinct().show()
+```
+
+Combining tile columns of different cell types gets a little trickier when user defined NoData cell types are involved. Let's create 3 tile columns: one without a defined NoData value, one with a NoData value of 1, and one with a NoData value of 2.
+
+
+```python
+def get_nodata_ct(nd_val):
```
vpipkt: again don't need to redefine this function

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,134 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+```
+
+When performing a local operation between tile columns with cell types `int` and type `float`, the resulting tile cell type will be `float`. If two tiles of different sized integer cell types are added together in the same way, the resulting cell type will be the same as the largest of the two integer cell types.
+
+```python
+rf_added = rf.withColumn('xy_sum', rf_local_add('y', 'x'))
+rf_added.select(rf_cell_type('xy_sum'), rf_cell_type('y'), rf_cell_type('x')).distinct().show()
+```
+
+Combining tile columns of different cell types gets a little trickier when user defined NoData cell types are involved. Let's create 3 tile columns: one without a defined NoData value, one with a NoData value of 1, and one with a NoData value of 2.
```
vpipkt: the first tile you should say one with the "default NoData value of zero" versus wihtout defined NoData. It's the distinction between `uint16` (has nodata, nodata value is 0) and `uint16raw` (does not have a nodata) .

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,134 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
...
+
+x = Tile((np.ones((100, 100))*3).astype('int16'))
+
+rf = spark.createDataFrame([Row(x=x)])
+
+rf_nd = rf.withColumn('x_nd_1', rf_convert_cell_type('x', get_nodata_ct(1))) \
+          .withColumn('x_nd_2', rf_convert_cell_type('x', get_nodata_ct(2)))
+```
+
+Let's try adding the tile column without a defined NoData value to a tile column with a defined NoData value. When there is an inconsistent NoData value in the two columns, the NoData value of the right-hand side of the sum is kept. In this case, this means the result has no defined NoData value.
```
vpipkt: may be worth a link to the local aglgebra page here.

vpipkt: again let's be careful with wording around "no defined NoData" versus "default NoData" ... if you're confused about it you can always use the CellType() methods `has_no_data` and `no_data_value`.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: Decided to add an explanation

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: The code in that comment can be used as the example in the documentation.

vpipkt:
@emc5ud when you get all your changes like you like, just ping me or re-request review

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -132,29 +135,134 @@ masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
 
 It's also nice to view a sample.
 
-```python show_masked
+```python
```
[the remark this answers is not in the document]
emc5ud: Right yeah I removed them because the text pyweave generates when calling display. I don't know the best way to handle that for now.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: My understanding is the `int16` does not have a NoData value of 0, while `uint16` does

emc5ud: Oh but that function makes tiles with `uint16` nvm

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: @vpipkt I added back the section with the local sum and NoData. I'd like you to check to make sure that sum is working correctly.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: @vpipkt I changed this section to remove the confusing between uint16 and int16.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
emc5ud: @vpipkt I believe I made the wording more clear.

PR #231, opened 2019-08-01
Refactor construction of LayoutDefinition in GeoTiff writing process

vpipkt:
This needs more unit tests. But in my little local check it seems to work correctly. Still more checks to do.

In particular I am concerned about a single moderate size input rasters  (500 x 800, 2000 x 2400) a single CRS, at full resolution, and round tripping them successfully.  Basically success meaning that the metadata agrees with the source and written data: crs, cell size (resolution), total size in cols and rows, extents. Would be interested in doing this for a bunch of different default (possibly non square also) tile dimensions (upon reading)


vpipkt:
If we don't get some kind of resolution soon, I would propose merging what we have here, removing teh assertion around addition, which is the only one that fails and only fails on Travis. And file an issue around investigating root cause and possibly fixing.

vpipkt:
Stacks on #236 please merge that first.

core/src/main/scala/org/locationtech/rasterframes/expressions/aggregates/ProjectedLayerMetadataAggregate.scala
```diff
@@ -76,7 +76,10 @@ class ProjectedLayerMetadataAggregate(destCRS: CRS, destDims: TileDimensions) ex
   override def evaluate(buffer: Row): Any = {
     import org.locationtech.rasterframes.encoders.CatalystSerializer._
     val buf = buffer.to[BufferRecord]
-    val LayoutLevel(_, layout) = FloatingLayoutScheme(destDims.cols, destDims.rows).levelFor(buf.extent, buf.cellSize)
+
+    val re = RasterExtent(buf.extent, buf.cellSize)
+    val layout = LayoutDefinition(re, destDims.cols, destDims.rows)
+
```
metasim: You'll have to fill me in on the deets; how you found this problem, the semantic difference, etc.

datasource/src/test/scala/org/locationtech/rasterframes/datasource/geotiff/GeoTiffDataSourceSpec.scala
```diff
@@ -101,6 +103,104 @@ class GeoTiffDataSourceSpec
...
+      val outputTif = SinglebandGeoTiff("target/unstructured_cog.tif")
+      outputTif.imageData.cols should be (inCols)
+      outputTif.imageData.rows should be (inRows)
+      outputTif.extent should be (inExtent)
+      outputTif.cellType should be (inCellType)
+
+    }
+
+    /*
+    it("should round trip jasons favorite unstructured raster round trip okay") {
```
metasim: keep?

pyrasterframes/src/main/python/docs/reference.pymd
```diff
@@ -238,6 +238,8 @@ If input `tile` had a NoData value already, the behaviour depends on if its cell
 
 [Local map algebra](https://gisgeography.com/map-algebra-global-zonal-focal-local/) raster operations are element-wise operations on a single tile (unary), between a `tile` and a scalar, between two `tile`s, or across many `tile`s.
 
+When these operations encounter a NoData value in either operand, the cell in the resulting `tile` will have a NoData.
```
metasim: 👍

pyrasterframes/src/main/python/tests/GeoTiffWriterTests.py
```diff
@@ -29,28 +29,34 @@ class GeoTiffWriter(TestEnvironment):
 
     @staticmethod
     def _tmpfile():
-        return os.path.join(tempfile.tempdir, "pyrf-test.tif")
+        return os.path.join(tempfile.gettempdir(), "pyrf-test.tif")
```
metasim: Is this a more stable way in Python?

datasource/src/test/scala/org/locationtech/rasterframes/datasource/geotiff/GeoTiffDataSourceSpec.scala
```diff
@@ -101,6 +103,104 @@ class GeoTiffDataSourceSpec
...
+      }
+      inCols should be (963)
+      inRows should be (754) //from gdalinfo
+      inExtent should be (Extent(752325.0, 3872685.0, 781215.0, 3895305.0))
+
+      val outputTif = SinglebandGeoTiff("target/unstructured_cog.tif")
+      outputTif.imageData.cols should be (inCols)
+      outputTif.imageData.rows should be (inRows)
+      outputTif.extent should be (inExtent)
+      outputTif.cellType should be (inCellType)
```
metasim: The GeoTrellis `testkit` includes `RasterMatchers` that maybe we could use to compare the actual values?

datasource/src/test/scala/org/locationtech/rasterframes/datasource/geotiff/GeoTiffDataSourceSpec.scala
vpipkt: removing that

core/src/main/scala/org/locationtech/rasterframes/expressions/aggregates/ProjectedLayerMetadataAggregate.scala
vpipkt: Was quite a bit of diggin around but the long and short was that the previous method snapped the resolution to the nearest zoom level, then returned the appropriate layout definition etc. This is more explicit.

pyrasterframes/src/main/python/tests/GeoTiffWriterTests.py
vpipkt: I am not 100% sure but i was having problems with `tempdir`

datasource/src/test/scala/org/locationtech/rasterframes/datasource/geotiff/GeoTiffDataSourceSpec.scala
vpipkt: Ah who knew! I looked around and they seemed mostly about making sure cell values across a tile match some condition, as opposed to these kinds of metadata equalities.

vpipkt:
@metasim i think this is ready to go. I'll wait for CI run and give you a chance to have another look

vpipkt:
Hmm travis keeps having time outs, either downloading artifacts or accessing raster data

PR #225, opened 2019-07-31
WIP: Python test_sql refactor all comparisons to inside SQL

vpipkt:
Hoping that this will pass Travis build.

PR #236, opened 2019-08-02
Fixed `Add` expression to use GT behavior

metasim:
Created new BiasedAdd for Aggregate use.

Addresses: https://github.com/locationtech/rasterframes/issues/229

Related: #225 #230

### DIFF
--- a/core/src/main/scala/org/locationtech/rasterframes/expressions/localops/BiasedAdd.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/expressions/localops/BiasedAdd.scala
@@ -20,7 +20,6 @@
  */
 
 package org.locationtech.rasterframes.expressions.localops
-
 import geotrellis.raster.Tile
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.codegen.CodegenFallback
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/locationtech/rasterframes/expressions/localops/BiasedAdd.scala
```diff
@@ -0,0 +1,74 @@
...
+  override protected def op(left: Tile, right: Double): Tile = DataBiasedOp.BiasedAdd(left, right)
+  override protected def op(left: Tile, right: Int): Tile = DataBiasedOp.BiasedAdd(left, right)
+
+  override def eval(input: InternalRow): Any = {
+    if(input == null) null
+    else {
+      val l = left.eval(input)
+      val r = right.eval(input)
+      if (l == null && r == null) null
+      else if (l == null) r
```
vpipkt: Hmm i am pretty lost with what this means all taken together.

metasim: @vpipkt It's for the case when one or the other are completely `null` Tiles.... no tiles.

vpipkt: not going to fall on my sword about it but a unit test around the intended behavior would be nice to see.

vpipkt:
@metasim See my latest commit. I'm pretty happy about the Travis CI passing (Whoop) 😅

FWIW Circle passes; `it:test` passes locally for me

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/BiasedAdd.scala
metasim: https://github.com/locationtech/rasterframes/issues/237

Issue #233, opened 2019-08-02
Python build zip for Maven


vpipkt:
`sbt pyrasterframes/package` prints something like this: 

```
[info] Maven Python .zip artifact written to '/Users/jbrown/src/raster-frames/pyrasterframes/target/scala-2.11/pyrasterframes-0.8.0-SNAPSHOT-py2.py3-none-any.whl'
```

But the statement does not make sense. Its saying the maven artifact is the whl? And I don't find a pyrasterframes zip anywhere in pyrasterframes/target after the command completes.

vpipkt:
try `sbt pyrasterframes/publishLocal` and check out the ~/.ivy2/local to see what would be published...

vpipkt:
Ok I am able to confirm that the wheel works just fine as far as i can tell

0) publish local : the source zip gets published to my local ivy caches as group/artifact/version/zips/name, in particular 

```
[info] Maven Python .zip artifact written to '/Users/jbrown/src/raster-frames/pyrasterframes/target/scala-2.11/pyrasterframes-0.8.0-SNAPSHOT-py2.py3-none-any.whl'
[info] :: delivering :: org.locationtech.rasterframes#pyrasterframes_2.11;0.8.0-SNAPSHOT :: 0.8.0-SNAPSHOT :: integration :: Thu Aug 08 13:35:01 EDT 2019
[info] 	delivering ivy file to /Users/jbrown/src/raster-frames/pyrasterframes/target/scala-2.11/ivy-0.8.0-SNAPSHOT.xml
[info] 	published pyrasterframes_2.11 to /Users/jbrown/.ivy2/local/org.locationtech.rasterframes/pyrasterframes_2.11/0.8.0-SNAPSHOT/docs/pyrasterframes_2.11-javadoc.jar
[info] 	published pyrasterframes_2.11 to /Users/jbrown/.ivy2/local/org.locationtech.rasterframes/pyrasterframes_2.11/0.8.0-SNAPSHOT/jars/pyrasterframes_2.11.jar
[info] 	published pyrasterframes_2.11 to /Users/jbrown/.ivy2/local/org.locationtech.rasterframes/pyrasterframes_2.11/0.8.0-SNAPSHOT/zips/pyrasterframes_2.11-python.zip
[...]

PR #241, opened 2019-08-05
Add prospective links to getting started docs

vpipkt:


pyrasterframes/src/main/python/docs/getting-started.pymd
```diff
@@ -1,6 +1,10 @@
 # Getting Started
 
-There are @ref:[several ways](getting-started.md#other-options) to use RasterFrames, and @ref:[several languages](languages.md) with which you can use it. Let's start with the simplest: the Python shell. Python 3.6 or greater is recommended.
+There are @ref:[several ways](getting-started.md#other-options) to use RasterFrames, and @ref:[several languages](languages.md) with which you can use it. Let's start with the simplest: the Python shell.  To get started you will need:
+
+1. [Python](https://www.python.org/) installed. Version 3.6 or greater is recommended.
+1. `pip` or `pip3` (recommended) installed. If you are using Python 3, `pip3` may already be installed.
+1. Java [JDK 8](https://openjdk.java.net/install/index.html) installed on your system and `java` on your system `PATH` or `JAVA_HOME` pointing to a Java installation.
```
vpipkt: Pretty sure this is true about java 8, but not totally sure.

pyrasterframes/src/main/python/docs/getting-started.pymd
```diff
@@ -43,15 +46,27 @@ You can also use RasterFrames in the following environments:
 
 ### Jupyter Notebook
 
-**TODO** User facing quick instructions e.g. how to pull and run docker hub hosted container
+ RasterFrames provides a Docker image for a Jupyter notebook server whose default kernel is already set up for running RasterFrames. To use it:
+
+ 1. Install [docker](https://docs.docker.com/install/)
+ 1. Pull the image: `docker pull s22s/rasterframes-notebook`
```
vpipkt: again pretty sure but not sure. locationtech also has a docker hub account. But not sure about what would be needed to publish there.

pyrasterframes/src/main/python/docs/getting-started.pymd
```diff
@@ -43,15 +46,27 @@ You can also use RasterFrames in the following environments:
...
+See [RasterFrames Notebook README](https://github.com/locationtech/rasterframes/blob/develop/rf-notebook/README.md) for instructions on building the Docker image for this Jupyter notebook server.
 
 ### `pyspark` shell or app
 
-To initialize RasterFrames in a `pyspark` shell, prepare to call pyspark with the appropriate `--master` and other `--conf` arguments for your cluster manager and environment. To these you will add the PyRasterFrames assembly JAR and the python source zip.
+To initialize RasterFrames in a `pyspark` shell, prepare to call pyspark with the appropriate `--master` and other `--conf` arguments for your cluster manager and environment. To these you will add the PyRasterFrames assembly JAR and the python source zip. 
+
+You can either [build](https://github.com/locationtech/rasterframes/blob/develop/README.md) the artifacts or download them:
+
+ * Assembly JAR: https://repo1.maven.org/maven2/org/locationtech/rasterframes/pyrasterframes_2.11/${VERSION}/pyrasterframes-assembly-${VERSION}.jar
```
vpipkt: This is a bit of a guess. I actually don't see the `assembly` jar published on maven central for the RC's

vpipkt: also note that the artifact name is missing the scala version.

metasim: Assemblies are *not* published on Maven Central... frowned upon.

We'll have to address the scala version tag later. Right now it's fine because it's not a published artifact, and gets added to the classpath outside the normal sbt chain.

pyrasterframes/src/main/python/docs/getting-started.pymd
```diff
@@ -43,15 +46,27 @@ You can also use RasterFrames in the following environments:
...
+ 1. In a browser, open `localhost:8808` in the example above. 
+
+See [RasterFrames Notebook README](https://github.com/locationtech/rasterframes/blob/develop/rf-notebook/README.md) for instructions on building the Docker image for this Jupyter notebook server.
 
 ### `pyspark` shell or app
 
-To initialize RasterFrames in a `pyspark` shell, prepare to call pyspark with the appropriate `--master` and other `--conf` arguments for your cluster manager and environment. To these you will add the PyRasterFrames assembly JAR and the python source zip.
+To initialize RasterFrames in a `pyspark` shell, prepare to call pyspark with the appropriate `--master` and other `--conf` arguments for your cluster manager and environment. To these you will add the PyRasterFrames assembly JAR and the python source zip. 
+
+You can either [build](https://github.com/locationtech/rasterframes/blob/develop/README.md) the artifacts or download them:
```
metasim: To be clear, while the assembly is not (nor will be) published on Maven Central, it is inside the PyPi .whl file, which will be published.

pyrasterframes/src/main/python/docs/getting-started.pymd
vpipkt: ah. so the instructions for this should read.. build with `sbt pyrasterframes/package` and then run this command?  

Or do we intend to publish the assembly jar in some other way?

vpipkt: Hmm yes i guess that's another option. User could unzip the whl file. Should we have them do that?

metasim: We will never publish the assembly outside of PyPi. Explain why we would need it?

metasim: BTW, once you publish to maven central, you can use RasterFrames like this:

```
spark-shell --packages org.locationtech.rasterframes:rasterframes_2.11:0.8.0-RC4
```

Spark launcher handles getting all the transitive dependencies.

metasim: If you want datasources:

```
org.locationtech.rasterframes:rasterframes-datasource_2.11:0.8.0-RC4
```

vpipkt: I think we are talking past each other a little. 

I *think* the zip sources artifact plus the `--packages` may be the way to go.  

But we need to fix #233 

```suggestion

pyspark \
    --conf spark.serializer=org.apache.spark.serializer.KryoSerializer \
    --conf spark.kryo.registrator=org.locationtech.rasterframes.util.RFKryoRegistrator \
    --conf spark.kryoserializer.buffer.max=500m \
    --py-files pyrasterframes_2.11-${VERSION}-python.zip \
    --packages org.locationtech.rasterframes:rasterframes_2.11:${VERSION},org.locationtech.rasterframes:pyrasterframes_2.11:${VERSION},org.locationtech.rasterframes:rasterframes-datasource_2.11:${VERSION} 

```

pyrasterframes/src/main/python/docs/getting-started.pymd
vpipkt: I think the `--packages` is better.

 see above.

vpipkt:
Depends on #233 fix at https://github.com/locationtech/rasterframes/pull/241/files#diff-a7b5e3e093a335c39112da2f76e4a151R71

vpipkt:
@metasim ready for another look i think, it should be about ready to go...

Issue #245, opened 2019-08-05
Overview Section Feedback


metasim:
1. You could add a sample DataFrame under "Context" section (after 1st paragraph). You have a sample DataFrame later on in the "Architecture" section, but maybe the same DataFrame without all the detail could be added here when you are introducing the DataFrame concept.
1. Graph supporting World Bank quote that shows expected explosion of data - some kind of line chart?
1. Why the lambda notation in the sample DataFrame in the tile columns?

vpipkt:
@courtney-whalen i'd be interested to chat about how we might address the first item, and what is driving that. 

I'm looking for somethign for the second one

ANd the third i'm trying to simplify.

vpipkt:
Addressing in #249

vpipkt:
@courtney-whalen @metasim 

Here is a sample of the rendered docs as on PR #249 . I've gotten rid of the many `/` that were inserted unhelpfully by my text editor. 

![image](https://user-images.githubusercontent.com/7798319/62625575-8043c700-b8f3-11e9-9b2b-48a37ac397f7.png)

courtney-whalen:
@vpipkt the first item isn't necessary. just a thought since that's where the DataFrame concept is introduced

Issue #247, opened 2019-08-05
Raster I/O Section Feedback


metasim:
1. I think more explanation is needed under Raster Writers section for each of those items.

vpipkt:
@courtney-whalen i plan to add links from each header to the writer page. I'll try to squeeze this into #249

Issue #246, opened 2019-08-05
Concepts Section Feedback


metasim:
1. I like the section on Cell Types - are these the only cell types that can be used in RasterFrames?
2. CRS:
  a. Is it useful to mention that GeoJSON files sometimes don't have a "crs" member and just default to 4326?
  b. I don't know how hard this would be to do, but some kind of visual that shows a single region reprojected to a couple different CRS and demonstrates how different projections distort the image would be interesting to me
3. It would be cool to have visuals showing scene, extent, tile, etc. and how they are related
4. The difference between extent and projected extent isn't totally clear to me
5. Projected raster - I don't think the "raster" concept is introduced yet. Might be useful to talk about that briefly first.

vpipkt:
Some very good ones here. Credit @courtney-whalen

I would tend to shy away from generating our own fundamental cartography/gis content like 2b, but it could be useful to try to find some links. This is maybe too basic but easy to understand and light hearted, no trig https://geoawesomeness.com/best-map-projection/

Our this is pretty good in term of principles and viz . https://gisgeography.com/map-distortion-tissots-indicatrix/

vpipkt:
I will break out item 3 it is a great idea but a heavier lift. Something quite similar was mentioned in #250

vpipkt:
Addressing the rest in #249

Issue #243, opened 2019-08-05
Some raster output in docs have lots of noise.


metasim:
From the "GeoTIFFs" section of _Writing Raster Data_:

<img width="460" alt="Screen Shot 2019-08-05 at 4 23 54 PM" src="https://user-images.githubusercontent.com/131013/62492747-6ee8a680-b79d-11e9-8bb8-65060e3058bc.png">

From the "User Defined Functions" section of _NumPy and Pandas_:

<img width="487" alt="Screen Shot 2019-08-05 at 4 25 07 PM" src="https://user-images.githubusercontent.com/131013/62492863-a0fa0880-b79d-11e9-8d74-7b87856e1d81.png">

I think this one is intended, but it's a bit confusing as an example, I think.

metasim:
If we have`rf_is_no_data` perhaps the second raster can be made more clear.

vpipkt:
FWIW this is what I see in the the Geoitiff section building locally.

The tiff itself seems correct


![image](https://user-images.githubusercontent.com/7798319/62626168-c4839700-b8f4-11e9-95a4-60a2052cc645.png)

vpipkt:
For what it's worth this is what i see in QGIS with no adjustments

Sort of typical for MODIS actually to have this kind of strange look to it if you don't do any adjustment to it.

![image](https://user-images.githubusercontent.com/7798319/62637098-d4a57180-b908-11e9-976c-5e6105d18ba8.png)

vpipkt:
We have cleaned up the first image in the latest doc version by only looking at a single channel.

Trying to stack three bands of modis to create a true color image is not that easily done  😉 

![image](https://user-images.githubusercontent.com/7798319/62729884-dc3e4680-b9ec-11e9-90cf-cc8d53676b9a.png)

PR #249, opened 2019-08-06
Feature/courtney edits

courtney-whalen:


pyrasterframes/src/main/python/setup.py
```diff
@@ -145,6 +145,7 @@ def initialize_options(self):
         'ipython==6.2.1',
         "ipykernel==4.8.0",
         'Pweave==0.30.3',
+        'fiona==1.8.6'
```
vpipkt: What was the reason for this? I thought fiona would come in with Shapely? see line 136?

But having said that I also had problems with shapely a few commits later.

courtney-whalen: I had an issue building the site because I had an outdated version of fiona, so Sim added this to prevent other people from having the same issue.

pyrasterframes/src/main/python/docs/raster-write.pymd
```diff
@@ -15,7 +15,7 @@ spark = pyrasterframes.get_spark_session()
 
 When collecting a _tile_ (see discussion of the RasterFrame @ref:[schema](raster-read.md#single-raster) for orientation to the concept) to the Python Spark driver, we have some convenience methods to quickly visualize the _tile_.
 
-In an IPython or Jupyter interpreter, a `Tile` object will be displayed as an image with limited metadata.
+In an IPython or Jupyter interpreter, a `Tile` object will be displayed as an image with limited metadata. In the example below, the displayed tile is mostly white due to cloud cover.
```
vpipkt: Umm see #243 i think there is now some issue in the generation of the image... the image is supposed to look like the first in the issue description, i think....

vpipkt: https://github.com/locationtech/rasterframes/issues/253

vpipkt: Filed https://github.com/locationtech/rasterframes/issues/253

vpipkt:
Fixes #245  #246 and #247 

Still open: #253 #252

Issue #169, opened 2019-06-28
RasterSource interpretation of file paths


vpipkt:
Assume we have a proper COG at /tmp/modis.tif.

`spark.read.rastersource('/tmp/modis.tif').count()`

Throws 

```
Error fetching data for [0,100000000e,646f6d2f706d742f,6669742e7369]
java.lang.UnsupportedOperationException: Reading '/tmp/modis.tif' not supported
```

Which is confusing and unhelpful.

The user meant

`spark.read.rastersource('file:///tmp/modis.tif').count()`

Can we 

1. Give a clearer error message that the uri was not understood, or
2. Look to see if the path starts with '/' and try prepending 'file://' ?

metasim:
We should do #2.

vpipkt:
I believe the current behavior is now that it won't throw an error but an empty dataframe is returned.

metasim:
🤦‍♂

vpipkt:
related... see docs/raster-write.pymd

PR #256, opened 2019-08-07
Use implicit file:// for schemeless paths with leading /

vpipkt:
Fix for #169  

The java.net.URI is a java object so possibly the URL.scheme can come back as `null` which was not always handled.

In the most important case in `RasterSource.IsDefaultGeoTiff.unapply` we made the faulty assumption that an empty scheme on a java.net.URI is the empty string `""` but never check for `null`.

Fixes the null check in that case.

This requires either no gdal support or config `rasterframes.prefer-gdal=false` to trigger it. Which is why it passed on circle but failed on travis.


vpipkt:
There is also some possible interplay with the `RasterSource.rsCache` at least when testing locally and such. I just dialed the cache timeout way down locally. Not sure how to overcome that.

vpipkt:
See #257 ...  travis build  failure actually helped me deliver better code ;-)

Issue #250, opened 2019-08-06
Reading Raster Data Page Feedback


courtney-whalen:
1. Would be cool to see the tile extents overlaid on the scene (if that's not too hard) when you talk about breaking the raster out into many arbitrary non-overlapping regions
	2. The URI formats table is all messed up in my site preview
	3. Lazy Raster Reads
		a. 2nd sentence trails off
		b. I think more explanation is needed on why lazy reading is useful
		c. Does rf_tile always request the realized tile from the lazy representation or just in the initial examples? Wording is a bit confusing here.
	4. Multiband Rasters
		a. How do you know which band is which index?

vpipkt:
For #1 we plan to  show something like this in the updated getting started jupyter notebook 

https://nbviewer.jupyter.org/github/locationtech/rasterframes/blob/b97582859256ca88416e39a1728d1cc9ed1a0901/rf-notebook/src/main/notebooks/Getting%20Started.ipynb 

It doesn't have the imagery also but... 🤷‍♂

vpipkt:
Reopening until #260 is merged. just for my sanity.

Issue #89, opened 2019-02-06
pyrasterframes read.geotrellis documentation


jdenisgiguere:
Greetings,

## Current situation

pyrasterframes expose a geotrellis reader. See:
https://github.com/locationtech/rasterframes/blob/e7209d312f7ce7af30a9c9c379532b2aff353e72/pyrasterframes/python/pyrasterframes/__init__.py#L54

I cannot find the syntax required to use it.

My current workaround is to use https://github.com/locationtech-labs/geopyspark/blob/d67f72c532d512f53eb18d9a0f2f8fd00ecced31/geopyspark/geotrellis/layer.py#L1235

## Expected situation

We should be able to use something like this.

```python
from pyspark.sql import SparkSession
from pyrasterframes import *

session = SparkSession.builder.appName("RasterFrameDemo").config("spark.ui.enabled", "false").getOrCreate().withRasterFrames()

rf = session.read.geotrellis({'uri': 'file:/data/my_catalog', 'layer_name': 'my_layer_name', 'layer_zoom': 0})

rf.printSchema()
```

jdenisgiguere:
As reported by @metasim , there is an alternative syntax.

**Geotrellis Layer**

```python
layer = spark.read.format('geotrellis').load(path=uri, layer=layer_name, zoom=zoom_level)
```
where
* `uri` : URI as string;
* `layer_name` : the layer name as string;
* `zoom_level` : the zoom level as integer.

**Geotrellis Catalog**

```python
catalog = spark.read.format('geotrellis-catalog').load(uri)
```
where
* `uri` : URI as string.

metasim:
Thanks @jdenisgiguere !!

Issue #252, opened 2019-08-06
Visual for scene vs tile extents


vpipkt:
Some feedback from #246 

>It would be cool to have visuals showing scene, extent, tile, etc. and how they are related

and #250 

> Would be cool to see the tile extents overlaid on the scene (if that's not too hard) when you talk about breaking the raster out into many arbitrary non-overlapping regions

vpipkt:
@metasim agree with milestone tag?

vpipkt:
bumping to milestone 0.8.1 because it is very much nice to have but don't want to hold us up. we can also do docs releases much faster than code releases.

vpipkt:
Here is a relatively quick way to do it.... Possibly we could ship code for this as part of the example notebooks in rf-notebook. [252_prototype.ipynb.zip](https://github.com/locationtech/rasterframes/files/3483831/252_prototype.ipynb.zip)


Rough outline

1. Select `rf_crs` and `rf_geometry` from rasterframe
2. toPandas
3. create geopandas GeoDataFrame
4. union all tile extents, stash away 
5. reproject original tif to mercator with rasterio **
6. Add geodataframe to folium map
7. Add mercator image (numpy array) to folium map (use union extents from 4)


![image](https://user-images.githubusercontent.com/7798319/62739306-7f4d8b00-ba02-11e9-84ca-7cf04ab5b3d2.png)



For production quality graphic we could probably just do the geometry overlay in QGIS with the original and take a screen capture at fairly high res...

For a more convincing interactive  in a notebook environment.
1. start with *small* and *mercator projected* raster ; spark.read.raster('/path/to/mercator.tif')  
2. rf.select(rf_crs('pr'), rf_tile('pr')).toPandas() 
3.  geopandas GeoDataframe
4. Add each extent shape and tile image to map using `GeoDataFrame.apply`
     - Get extent in mercator, convert to WGS84 / EPSG4326
     - map.addChild(folium.GeoJson(extent_4326))
     - map.addChile(folium.raster_layers.ImageOverlay(image=tile.cells, bounds=extent_4326))

vpipkt:
@courtney-whalen ☝️

courtney-whalen:
looking good! this is what I had in mind for that visual

vpipkt:
IMplementing this in docs would be aided by #302

vpipkt:
Related #342

vpipkt:
[...]

PR #260, opened 2019-08-07
Docs fixes in raster read 

vpipkt:
#250 #89

### DIFF
--- a/pyrasterframes/src/main/python/docs/raster-read.pymd
+++ b/pyrasterframes/src/main/python/docs/raster-read.pymd
@@ -193,7 +193,7 @@ mb2.printSchema()
 A GeoTrellis catalog is a set of GeoTrellis layers. We can read a dataframe giving details of the content of a catalog using the following. The scheme is typically `hdfs` or a cloud storage provider like `s3` or `wasb`.
 
 ```python, evaluate=False
-gt_cat = spark.read.format('geotrellis-catalog').load('scheme://path-to-gt-catalog')
+gt_cat = spark.read.geotrellis_catalog('scheme://path-to-gt-catalog')
 ```
 
 ### GeoTrellis Layers
--- a/pyrasterframes/src/main/python/docs/raster-read.pymd
+++ b/pyrasterframes/src/main/python/docs/raster-read.pymd
@@ -8,11 +8,13 @@ from pyrasterframes.rasterfunctions import *
 spark = create_rf_spark_session()
 ```
 
-RasterFrames registers a DataSource named `raster` that enables reading of GeoTIFFs (and other formats when @ref:[GDAL is installed](getting-started.md#installing-gdal)) from arbitrary URIs. In the examples that follow we'll be reading from a Sentinel-2 scene stored in an AWS S3 bucket.
+RasterFrames registers a DataSource named `raster` that enables reading of GeoTIFFs (and other formats when @ref:[GDAL is installed](getting-started.md#installing-gdal)) from arbitrary URIs. The `raster` DataSource operates on either a single raster file location or another DataFrame, called a _catalog_, containing pointers to many raster file locations.
+
+RasterFrames can also read from @ref:[GeoTrellis catalogs and layers](raster-read.md#geotrellis).
 
 ## Single Raster
 
-The simplest form is reading a single raster from a single URI.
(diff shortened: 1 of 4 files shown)


vpipkt:
@courtney-whalen take a quick look please let us know if you think it handles #250 less already logged #252

courtney-whalen:
@vpipkt just reviewed - looks good to me! I closed #250.

pyrasterframes/src/main/python/docs/raster-read.pymd
```diff
@@ -56,19 +58,19 @@ RasterFrames relies on three different IO drivers, selected based on a combinati
...
+| `file://`           | yes         | yes      | no       |
+| `http://`           | yes         | yes      | no       |
+| `https://`          | yes         | yes      | no       |
+| `ftp://`            | `/vsicurl/` | yes      | no       |
+| `hdfs://`           | `/vsihdfs/` | no       | yes      |
+| `s3://`             | `/vsis3/`   | yes      | no       |
+| `s3n://`            | no          | no       | yes      |
+| `s3a://`            | no          | no       | yes      |
+| `wasb://`           | `/vsiaz/`   | no       | yes      |
+| `wasbs://`          | no          | no       | yes      |
```
metasim: 👍 FYI: The unicode characters caused problems with (LaTeX) PDF generation.

pyrasterframes/src/main/python/docs/raster-read.pymd
```diff
@@ -142,15 +144,17 @@ spark.read.raster(uri, lazy_tiles=False) \
...
 
 ## Multiband Rasters
 
-A multiband raster represents a three dimensional numeric array. The first two dimensions are spatial, and the third dimension is typically designated as different bands. The bands could represent intensity of different wavelengths of light (or other electromagnetic radiation), or they could represent other phenomena such as measurement time, quality indications, or additional measurements.
+A multiband raster represents a three dimensional numeric array. The first two dimensions are spatial, and the third dimension is typically designated as different bands. The bands could represent intensity of different wavelengths of light (or other electromagnetic radiation), or they could represent other phenomena such as measurement time, quality indications, or additional measurements. 
+
+Multiband rasters files have a strictly ordered set of bands, which are typically indexed from 1. Some files have metadata tags associated with each band. Some files have a color interpetation metadata tag indicating how to interpret the bands.
```
metasim: > which are typically indexed from 1

Do we need to be careful here in indicating that "band index" specifications in RasterFrames are zero based?

metasim: I see that you point that out below, but I think further clarification that one-based is "colloquially" or "conventionally" in the context of EO dialog.

pyrasterframes/src/main/python/docs/raster-read.pymd
```diff
@@ -180,6 +184,29 @@ mb2 = spark.read.raster(catalog=spark.createDataFrame(mb_cat),
...
+## GeoTrellis
+
+### GeoTrellis Catalogs
+
+[GeoTrellis][GeoTrellis] is one of the key libraries that RasterFrames builds upon. It provides a Scala language API to working with large raster data with Apache Spark. RasterFrames provides a DataSource that supports both reading and @ref:[writing](raster-write.md#geotrellis-layers) with GeoTrellis.
+
+A GeoTrellis catalog is a set of GeoTrellis layers. We can read a dataframe giving details of the content of a catalog using the following. The scheme is typically `hdfs` or a cloud storage provider like `s3` or `wasb`.
+
+```python, evaluate=False
+gt_cat = spark.read.format('geotrellis-catalog').load('scheme://path-to-gt-catalog')
```
metasim: We should create an extension method for this to be consistent with other usesages.

pyrasterframes/src/main/python/docs/raster-read.pymd
vpipkt: yes and i think the use of `+` and `-` broke the table formatting in paradox or someplace

pyrasterframes/src/main/python/docs/raster-read.pymd
vpipkt: My understanding of GDAL anyways is that they are defined as one-based index: 

https://gdal.org/doxygen/classGDALDataset.html#ad96adcf07f2979ad176e37a7f8638fb6 

My concern is the user that's a bit familiar with EO data does a `gdalinfo` and sees band 1 and 2 are what they need, then put `band_indexes=(1,2)`  


Maybe I'm missing your point?

Issue #238, opened 2019-08-05
Updated functions accepting a cell type to allow `CellType` class.


metasim:
@vpipkt We should fix this to accept the `CellType` class directly, and not have the user do `cell_type_name`.

_Originally posted by @metasim in https://github.com/locationtech/rasterframes/pull/230/files_

vpipkt:
I've contributed to #230 

I will add some more unit tests to it once that branch merges develop back in.

vpipkt:
Fixed. and demonstrated in docs. could still use a unit test or two.

PR #271, opened 2019-08-08
Document and improve use of CellType in rasterfunctions args

vpipkt:
Note the more careful parsing of user defined nodata.

It's tagged 0.8.0 because it is a small fry but we can easily push it off.

### DIFF
--- a/pyrasterframes/src/main/python/docs/reference.pymd
+++ b/pyrasterframes/src/main/python/docs/reference.pymd
@@ -126,26 +126,29 @@ Functions to create a new Tile column, either from scratch or from existing data
 ### rf_make_zeros_tile
 
 ```
-Tile rf_make_zeros_tile(Int tile_columns, Int tile_rows, String cell_type_name)
+Tile rf_make_zeros_tile(Int tile_columns, Int tile_rows, [CellType cell_type])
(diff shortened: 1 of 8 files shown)


vpipkt:
To close #238

pyrasterframes/src/main/python/docs/reference.pymd
```diff
@@ -126,26 +126,29 @@ Functions to create a new Tile column, either from scratch or from existing data
 ### rf_make_zeros_tile
 
 ```
-Tile rf_make_zeros_tile(Int tile_columns, Int tile_rows, String cell_type_name)
+Tile rf_make_zeros_tile(Int tile_columns, Int tile_rows, [CellType cell_type])
+Tile rf_make_zeros_tile(Int tile_columns, Int tile_rows, [String cell_type_name])
```
metasim: @vpipkt Sometime before 1.0, we should change these signatures to use the `var: Type` ordering, as that's what type hints in Python 3.7 use (along with Scala).

That said, really want to get this stuff generated by Sphinx.

vpipkt:
@metasim the function to get celltype had already been refactored, this PR was just clean up https://github.com/locationtech/rasterframes/pull/271/files#diff-7a654590ae01cfdf224abac0aa84caeeL211-L212

PR #275, opened 2019-08-09
Tile render adjustments

emc5ud:
A few changes:

- Tile dimensions no longer affect ipython render size. 10x10 tile renders the same size as a 1000x1000

- Tile columns render such that the tiles are slightly smaller than if the tile was rendered by itself. 

- Font size of the title scales with the render size of the tile.

### DIFF
--- a/pyrasterframes/src/main/python/pyrasterframes/rf_ipython.py
+++ b/pyrasterframes/src/main/python/pyrasterframes/rf_ipython.py
@@ -21,7 +21,7 @@
 import pyrasterframes.rf_types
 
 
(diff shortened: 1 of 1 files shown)


emc5ud:
image
image
image

emc5ud:
@vpipkt

metasim:
@emc5ud Nice work. Don't forget to select one or more "Reviewers" in the upper right.

emc5ud:
@metasim Is there a process that gives me permissions to do that? Right now the best I can do is tag someone in a comment

PR #284, opened 2019-08-14
Ability to render pymd 'show' tables to Markdown .

metasim:

PR #286, opened 2019-08-14
Better tile rendering

emc5ud:

Default tile rendering is now more like QGIS. By default, the 1st and 99th percentile are calculated for the image, and then the cell values are clipped to not go beyond this range. 

Before:
image
After:
image


emc5ud:
TODO: I'd like to have a `show` method that makes it easy to pass in rendering arguments

emc5ud:
Ok so I changed my methodology a little bit and would love some feedback from @metasim @vpipkt  and/or @courtney-whalen. 

I added a method to the Tile class called `show` that normalizes the image and uses `matplotlib.pyplot.imshow` to display. This function will either create a new axis object to put the image onto or you can pass in a premade axis object you want to modify. 

`tile_to_png` now calls the `Tile.show` method to create an image with a prespecified size and title.

The existing behavior of tile_to_png is unchanged, but this gives an easy way to do more custom tile plots. For example

image

`tile.show` has full flexibility to pass in any argument someone would pass into `matplotlib.pyplot.imshow`

vpipkt:
Based on the quick look I like it and I think it's going to be a good contribution.

Can you take a look at possibly rebasing your work off of the branch in  #284 as it also makes some changes in the same module?

metasim:
So far I like it. I made some (not as good as) changes the same code, so there'll be a merge conflict when you update to the latest `develop`. I also may want to consider where `show` actually lives.

metasim:
@emc5ud PS: **definitely** looks better!!! Thank you!!

emc5ud:
Thanks for taking a look. I might need some help with rebasing this branch off the one you linked.

PR #294, opened 2019-08-19
Added Markdown and HTML rendering of Spark DataFrames

metasim:


pyrasterframes/src/main/python/setup.py
```diff
@@ -104,7 +104,7 @@ def run(self):
             if (not self.quick) or (not path.exists(dest)) or (path.getmtime(dest) < path.getmtime(file)):
                 print(_divided('Running %s' % name))
                 try:
-                    pweave.weave(file=str(file), doctype=self.format, cache=True)
+                    pweave.weave(file=str(file), doctype=self.format)
```
metasim: This had unintended consequences.

core/src/main/scala/org/locationtech/rasterframes/util/package.scala
```diff
@@ -203,6 +203,25 @@ package object util {
...
     }
+
+    def toHTML(numRows: Int = 5, truncate: Boolean = false): String = {
+      import df.sqlContext.implicits._
+      val cols = df.columns
+      val header = "<thead>\n" + cols.mkString("<tr><th>", "</th><th>", "</th></tr>\n") + "</thead>\n"
+      val stringifiers = cols
+        .map(c => s"`$c`")
+        .map(c => df.col(c).cast(StringType))
+        .map(c => if (truncate) substring(c, 1, 40) else c)
```
vpipkt: if truncate is true, would like to see a representation that the string has been trunctated. E.g. `Some really really long string about something complicated`

> `Some really really long string about som` 👎 
> `Some really really long string about ...` 👍 

This would be the same in the `toMarkdown` method.

pyrasterframes/src/main/python/pyrasterframes/rf_types.py
```diff
@@ -358,11 +358,6 @@ def dimensions(self):
         """ Return a list of cols, rows as is conventional in GeoTrellis and RasterFrames."""
         return [self.cells.shape[1], self.cells.shape[0]]
 
-    def _repr_png_(self):
-        """Provide default PNG rendering in IPython and Jupyter"""
-        from pyrasterframes.rf_ipython import tile_to_png
-        return tile_to_png(self)
-
```
metasim: This is now handed via `ip.display_formatter.formatters['image/png']`

pyrasterframes/src/main/python/pyrasterframes/rf_ipython.py
```diff
@@ -167,20 +173,26 @@ def spark_df_to_markdown(df, num_rows=5, truncate=True, vertical=False):
...
+        # Can be removed if we can figure out a way to pass settings through `display`
+        pyspark.sql.DataFrame.showMarkdown = spark_df_to_markdown
+        pyspark.sql.DataFrame.showHTML = spark_df_to_html
+        Tile.show = plot_tile
 
-        # See if we're in documentation mode and register a custom show implementation.
-        if 'InProcessInteractiveShell' in ip.__class__.__name__:
-            pyspark.sql.DataFrame._repr_markdown_ = spark_df_to_markdown
-            pyspark.sql.DataFrame.show = lambda df, num_rows=5, truncate=True: display_markdown(spark_df_to_markdown(df, num_rows, truncate), raw=True)
+        # This is a trick we may have to introduce above.
```
metasim: This didn't really work, and as it turns out, registering a separate Markdown vs. HTML renderer takes care of the difference between documentation generation vs. notebooks.

core/src/main/scala/org/locationtech/rasterframes/util/package.scala
vpipkt: See my prior mention of this... https://github.com/locationtech/rasterframes/pull/284#discussion_r314414198

pyrasterframes/src/main/python/docs/aggregation.pymd
```diff
@@ -33,14 +33,16 @@ print(tiles[1]['tile'].cells)
 
 We use the @ref:[`rf_tile_mean`](reference.md#rf-tile-mean) function to compute the _tile_ aggregate mean of cells in each row of column `tile`. The mean of each _tile_ is computed separately, so the first mean is 1.0 and the second mean is 3.0. Notice that the number of rows in the DataFrame is the same before and after the aggregation.
 
-```python, tile_mean, results='raw'
-rf.select(F.col('id'), rf_tile_mean(F.col('tile'))).show()
+```python, tile_mean
+means = rf.select(F.col('id'), rf_tile_mean(F.col('tile')))
+display(means)
```
vpipkt: i believe it may not be necesarry to expliclty call display? At least from my attempts to build the docs locally.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
```diff
@@ -135,14 +139,15 @@ Now we will use the @ref:[`rf_mask_by_value`](reference.md#rf-mask-by-value) to
...
+    .drop('nodata', 'defect', 'cloud8', 'cloud9', 'cirrus', 'blue')
 ```
 
 We can verify that the number of NoData cells in the resulting `blue_masked` column matches the total of the boolean `mask` _tile_ to ensure our logic is correct.
 
-```python, show_masked, results='raw'
-masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask')).show(10)
+```python, show_masked
+counts = masked.select(rf_no_data_cells('blue_masked'), rf_tile_sum('mask'))
+display(counts)
```
vpipkt: In some of these examples if we can just have a spark DF on the last line that will have `display` evaluated on  it, it would improve readability some to revert back.

core/src/main/scala/org/locationtech/rasterframes/util/package.scala
```diff
@@ -203,6 +203,25 @@ package object util {
...
+      val cat = concat_ws("</td><td>", stringifiers: _*)
+      val body = df
+        .select(cat).limit(numRows)
```
vpipkt: So we would really like to have some indication on the output that truncation has happened. Even if it is just like teh spark show message `"Showing only top 5 rows."` that would be better....

Obviously we could have a case where we print that and there is only 1 row or the DF is empty.

Possibly collect then check if length is < `numRows` to adjust appearance of message?

core/src/main/scala/org/locationtech/rasterframes/util/package.scala
```diff
@@ -203,6 +203,25 @@ package object util {
         .mkString("| ", " |\n| ", " |")
       header + body
     }
+
+    def toHTML(numRows: Int = 5, truncate: Boolean = false): String = {
+      import df.sqlContext.implicits._
+      val cols = df.columns
+      val header = "<thead>\n" + cols.mkString("<tr><th>", "</th><th>", "</th></tr>\n") + "</thead>\n"
+      val stringifiers = cols
```
vpipkt: it may also make sense to have some truncation parameter /  check on the number of Columns? In some ML applications there can be hundreds of columns which are not going to be meaningful to look at. 

As below if truncating need to give some kind of message / visual indication. Pandas does this well.

pyrasterframes/src/main/python/docs/nodata-handling.pymd
metasim: I like that idea... not sure I realized that behavior at the time.

core/src/main/scala/org/locationtech/rasterframes/util/package.scala
metasim: Done :-)

core/src/main/scala/org/locationtech/rasterframes/util/package.scala
metasim: You'll have to explain to me what kind of assumptions you can make about what's meaningful or not.

PR #298, opened 2019-08-19
Removed typed column specification for ProjectedRasterTile functions.

### LINKED ISSUE #295: Dynamically typed ProjectedRasterTile expressions don't deserialize to Tile
If `minus` and `one` are `ProjectedRasterTile`s, the following code will throw a runtime cast exception.

```scala
val df = Seq((minus, one)).toDF("minus", "one")
assertEqual(df.select(rf_abs($"minus")).first(), one)
```

error:

```console
cannot resolve 'CAST(`rf_abs(minus)` AS STRUCT<`cell_context`: STRUCT<`cellType`: STRUCT<`cellTypeName`: STRING>, `dimensions`: STRUCT<`cols`: SMALLINT, `rows`: SMALLINT>>, `cell_data`: STRUCT<`cells`: BINARY, `ref`: STRUCT<`source`: STRUCT<`raster_source_kryo`: BINARY>, `bandIndex`: INT, `subextent`: STRUCT<`xmin`: DOUBLE, `ymin`: DOUBLE, `xmax`: DOUBLE, `ymax`: DOUBLE>>>>)' due to data type mismatch: cannot cast struct<tile_context:struct<extent:struct<xmin:double,ymin:double,xmax:double,ymax:double>,crs:struct<crsProj4:string>>,tile:tile> to tile;
```

metasim:
Closes #295

### DIFF
--- a/core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
@@ -68,6 +68,7 @@ package object expressions {
     registry.registerExpression[ExplodeTiles]("rf_explode_tiles")
     registry.registerExpression[GetCellType]("rf_cell_type")
     registry.registerExpression[SetCellType]("rf_convert_cell_type")
+    registry.registerExpression[SetNoDataValue]("rf_with_nodata")
     registry.registerExpression[GetDimensions]("rf_dimensions")
     registry.registerExpression[ExtentToGeometry]("st_geometry")
     registry.registerExpression[GetGeometry]("rf_geometry")
--- a/core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
@@ -68,7 +68,7 @@ package object expressions {
     registry.registerExpression[ExplodeTiles]("rf_explode_tiles")
     registry.registerExpression[GetCellType]("rf_cell_type")
     registry.registerExpression[SetCellType]("rf_convert_cell_type")
-    registry.registerExpression[SetNoDataValue]("rf_with_nodata")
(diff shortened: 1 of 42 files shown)


core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
```diff
@@ -68,6 +68,7 @@ package object expressions {
     registry.registerExpression[ExplodeTiles]("rf_explode_tiles")
     registry.registerExpression[GetCellType]("rf_cell_type")
     registry.registerExpression[SetCellType]("rf_convert_cell_type")
+    registry.registerExpression[SetNoDataValue]("rf_with_nodata")
```
vpipkt: naming, this is otherwise referred to as `rf_with_no_data`.

core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/SetCellType.scala
```diff
@@ -21,43 +21,52 @@
...
+
+@ExpressionDescription(
+  usage = "_FUNC_(tile, value) - Set the cell type for the given tile.",
+  arguments = """
+  Arguments:
+    * tile - left-hand-side tile
+    * rhs  - a cell type definition""",
+  examples = """
+  Examples:
+    > SELECT _FUNC_(tile, 1.5);
```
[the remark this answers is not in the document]
metasim: oops.

core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
metasim: The names don't always match up, and the thought of changing it to `WithNoData` doesn't sit well with me when I look at the other class names. I'm not sure I'm a fan of `rf_convert_cell_type` or `rf_with_nodata`, but going and making everything precisely consistent seems like a rabbit hole. I'd like to leave it if it's OK with you.

Issue #285, opened 2019-08-14
Figure out how to render geojson in time-series.pymd


metasim:
Folium works in the notebook... can we make it work for the documentation? At a minimum, we could capture the output from the notebook and insert it as static HTML content.

vpipkt:
fun fact, simplifying the geometry is probably smart. Here is a quick look at the simplify param versus number of vertices in the multipolygon:

```
0 		 9217
1e-07 		 4711
1e-06 		 2908
1e-05 		 1568
0.0001 		 921
0.001 		 320
```

Those are in degrees so 0.001 is nominally 110m; Modis pixel size is ~465m so that's probably about appropriate.

vpipkt:
In notebook the `display` of the shapely geometry object has a `_repr_svg_` method so

![image](https://user-images.githubusercontent.com/7798319/63277826-0a851700-c274-11e9-8ed4-5ac008ab0aca.png)

vpipkt:
So far I'm just getting the cell output of the display object...

`<IPython.display.GeoJSON object>`

metasim:
@vpipkt I wonder what would happen if you also registered that formatter as the handler for `text/markdown`... since injecting HTML into markdown often works when the output target is HTML, it may get passed through and work as expected.

vpipkt:
💥 good idea. see also issue #302

vpipkt:
Immediate issue fixed by #299 more improvements coming with #302

Issue #283, opened 2019-08-14
Update time-series.pymd to use geojson spark reader.


metasim:
If we prep the data so it's valid geojson then we can use the built-in reader and reduce ceremony.

PR #299, opened 2019-08-19
Time series docs improvements 

### LINKED ISSUE #288: Update time-series.pymd to only use DataFrame API
This will help tease out problems with literal/column type mismatching in Python.

vpipkt:
In the time-series documentation page:

* configure circle CI to save intermediate markdown files as artifacts on docs build failure
* simplify rest request to argis
* rest request specify `geojson` format
* attempt to embed folium
   * generate folium.Map
   * save it as html to temp location
   * create an `<iframe>` within a markdown file to `@@include`
* read geojson with spark DataSource
* simplify geometry
* cross join catalog and vector data
* try to speed up execution?

Should fix #288 #283 #285 

Please squash merge!

### DIFF
--- a/pyrasterframes/src/main/python/docs/time-series.pymd
+++ b/pyrasterframes/src/main/python/docs/time-series.pymd
@@ -92,7 +92,7 @@ park_cat.printSchema()
 
 Now we have a catalog with several months of MODIS data for a single granule. However, the granule is larger than our park boundary. We will combine the park geometry with the catalog, and read only the bands of interest to compute NDVI, which we discussed in a @ref:[previous section](local-algebra.md#computing-ndvi).
 
-We then [reproject](https://gis.stackexchange.com/questions/247770/understanding-reprojection) the park geometry to the same @ref:[CRS][concepts.md#coordinate-reference-system--crs-] as the imagery. Then we will filter to only the _tiles_ intersecting the park.
+We then [reproject](https://gis.stackexchange.com/questions/247770/understanding-reprojection) the park geometry to the same @ref:[CRS](concepts.md#coordinate-reference-system--crs-) as the imagery. Then we will filter to only the _tiles_ intersecting the park.
 
 ```python read_catalog
 raster_cols = ['B01', 'B02',] # red and near-infrared respectively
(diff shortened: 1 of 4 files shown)


vpipkt:
Rendered docs build, highlight: 

For #285 : https://1258-174014757-gh.circle-artifacts.com/0/rf-site/time-series.html#vector-data

The code for the folium map embedding could be wrapped up in a function. It would also be really nicer to have a cell output rather than relying on the @@include.

pyrasterframes/src/main/python/docs/time-series.pymd
```diff
@@ -33,16 +33,17 @@ First we will get the vector data from LandsNet service by a REST query. The dat
 
 ```python get_park_boundary
 import requests
+nps_filepath = '/tmp/parks.geojson'
 nps_data_query_url = 'https://services1.arcgis.com/fBc8EJBxQRMcHlei/arcgis/rest/services/' \
                      'NPS_Park_Boundaries/FeatureServer/0/query' \
                      '?geometry=-82.451,41.075,-80.682,41.436&inSR=4326&outSR=4326&f=geojson'
 r = requests.get(nps_data_query_url)
```
metasim: Did you try doing this with `SparkFiles` like this?

https://github.com/locationtech/rasterframes/blob/bdcfb39a4c8c7276e9a6c8d0ae41d81fd31f54df/pyrasterframes/src/main/python/docs/vector-data.pymd#L15-L18

Not requiring the use of `requests` (even though it is a nice library) would clear a bit more cognitive load.

metasim: Just saw the comments in #288 which seem to indicate you tried the SparkFiles thing. I suspect the problem is the query parameter?

metasim:
@vpipkt Can you squash merge the few character commit messages?

metasim:
> Rendered docs build, highlight:
> 
> For #285 : https://1258-174014757-gh.circle-artifacts.com/0/rf-site/time-series.html#vector-data
> 
> The code for the folium map embedding could be wrapped up in a function. It would also be really nicer to have a cell output rather than relying on the @@include.

I agree, but even this is a significant improvements... think we can make it taller? Note: if css hacks are needed you can drop something in `docs/src/main/paradox/_template/page.st`.

pyrasterframes/src/main/python/docs/time-series.pymd
vpipkt: The problem seems to be that SparkFiles does not know what the filename should be, but it is opaque actually. The call to "Add" the file seems to work or fail silently. But reading it doesn't and you can kind of peek at the local cache to see nothing is there.

PR #300, opened 2019-08-20
Add unit test for rf_explode_tiles_resample

### LINKED ISSUE #76: Python binding for explode Tiles with sampling
Sometimes exploding is slow and we just want to see a little bit of the data... in Python.

vpipkt:
This branch was lingering around but #77 is merged and supposed to close #76 .... wonder if this test passes?


vpipkt:
Maybe we could reopen issue around sql registration for `rf_explode_tile_sample`? seems possible but in a quick exploration i could not get it quite right...

PR #301, opened 2019-08-20
Adds rf_render_png and rf_rgb_composite.

metasim:


pyrasterframes/src/main/python/docs/numpy-pandas.pymd
```diff
@@ -32,15 +32,15 @@ t.cells.shape, t.cells.nbytes
 
 As discussed in the @ref:[raster writing chapter](raster-write.md#dataframe-samples), a pretty display of Pandas DataFrame containing _tiles_ is available by importing the `rf_ipython` submodule. In addition, as discussed in the @ref:[vector data chapter](vector-data.md), any geometry type in the Spark DataFrame will be converted into a Shapely geometry. Taken together, we can easily get the spatial information and raster data as a NumPy array, all within a Pandas DataFrame.
 
-```python spark_session,  echo=False
+```python, spark_session,  echo=False
```
vpipkt: this page could use a little rethinking. The user should prefer to `display(df)` than to do `display(df.toPandas)` which is how it's still written.

pyrasterframes/src/main/python/docs/raster-write.pymd
```diff
@@ -35,26 +35,21 @@ display(tile) # IPython.display function
 
 ## DataFrame Samples
 
-Within an IPython or Jupyter interpreter, a Pandas DataFrame containing a column of _tiles_ will be rendered as the samples discussed above. Simply import the `rf_ipython` submodule to enable enhanced HTML rendering of a Pandas DataFrame.
+Within an IPython or Jupyter interpreter, a Spark DataFrame containing a column of _tiles_ will be rendered as the samples discussed above. Simply import the `rf_ipython` submodule to enable enhanced HTML rendering of a Spark DataFrame.
```
vpipkt: It is true for either Spark or Pandas. We should mention that here. And "The rendering for Spark DataFrames will be more efficient than converting a Spark DataFrame to Pandas.

pyrasterframes/src/main/python/docs/reference.pymd
```diff
@@ -10,6 +10,15 @@ ReturnDataType function_name(InputDataType argument1, InputDataType argument2)
 
 For the Scala documentation on these functions, see @scaladoc[`RasterFunctions`][RasterFunctions]. The full Scala API documentation can be found [here][scaladoc].
 
+```python imports, echo=False
```
vpipkt: not sure we have anything that runs any more? possibly this file can be converted to `md` instead of `pymd`

pyrasterframes/src/main/python/docs/reference.pymd
```diff
@@ -608,5 +617,19 @@ Pretty print the tile values as plain text.
 
 Render Tile cell values as numeric values, for debugging purposes.
 
+
+### rf_rgb_composite
+
+    Tile rf_rgb_composite(Tile red, Tile green, Tile blue)
+    
+Merges three bands into a single byte-packed RGB composite. It first scales each cell to fit into an unsigned byte, in the range 0-255, and then merges all three channels to fit into a 32-bit unsigned integer.
```
vpipkt: Would be well worth a mention as to why the user would want to do that?

core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
```diff
@@ -128,6 +128,9 @@ package object expressions {
 
     registry.registerExpression[DebugRender.RenderAscii]("rf_render_ascii")
     registry.registerExpression[DebugRender.RenderMatrix]("rf_render_matrix")
+    registry.registerExpression[RenderPNG.RenderCompositePNG]("rf_render_png")
```
vpipkt: So this guy only takes one param but it is documented (in reference.pymd) as if it takes 3. Not sure what the intent is. I would think we should have one overload for composite PNG and another overload for ColorRamp PNG.

pyrasterframes/src/main/python/pyrasterframes/rasterfunctions.py
```diff
@@ -348,6 +362,16 @@ def rf_render_matrix(tile_col):
     return _apply_column_function('rf_render_matrix', tile_col)
 
 
+def rf_render_png(red_tile_col, green_tile_col, blue_tile_col):
```
vpipkt: we should have a single tile overload or another function to do the ColorRamp

core/src/main/scala/org/locationtech/rasterframes/RasterFunctions.scala
```diff
@@ -81,6 +83,10 @@ trait RasterFunctions {
   def rf_assemble_tile(columnIndex: Column, rowIndex: Column, cellData: Column, tileCols: Int, tileRows: Int, ct: CellType): TypedColumn[Any, Tile] =
     rf_convert_cell_type(TileAssembler(columnIndex, rowIndex, cellData, lit(tileCols), lit(tileRows)), ct).as(cellData.columnName).as[Tile](singlebandTileEncoder)
 
+  /** Create a Tile from a column of cell data with location indexes and preform cell conversion. */
```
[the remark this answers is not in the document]
metasim: > We should also revisit the notebook about rendering things. We may be able to remove the notebook if we do as you mentioned creating a page about exploring the data?

I'd like to have the rendered docs be fully complete. Another way of thinking about it is we could have a dedicated "Jupyter" chapter.

core/src/main/scala/org/locationtech/rasterframes/ref/RasterSource.scala
```diff
@@ -61,7 +61,7 @@ trait RasterSource extends ProjectedRasterLike with Serializable {
     readBounds(Seq(bounds), bands).next()
 
   def read(extent: Extent, bands: Seq[Int] = SINGLEBAND): Raster[MultibandTile] =
-    read(rasterExtent.gridBoundsFor(extent, clamp = true), bands)
+    read(rasterExtent.gridBoundsFor(extent, clamp = false), bands)
```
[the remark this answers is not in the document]
metasim: I'll switch it back... I was seeing if it made any difference and it doesn't, but shouldn't have kept it. Yeh, cruft.. thanks.

metasim:
@vpipkt 

> We should also revisit the notebook about rendering things. We may be able to remove the notebook if we do as you mentioned creating a page about exploring the data?

I'd like to have the rendered docs be fully self-contained/complete, so I'm in favor of that being a goal. 

Another way of thinking about it is we could have a dedicated "Jupyter" chapter.

vpipkt:
Another recommended change. We should truncate the WKT representation of various geometries on the vector-data page. They are extremely long. 

image

core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
metasim: #313

pyrasterframes/src/main/python/pyrasterframes/rasterfunctions.py
metasim: #313

PR #322, opened 2019-08-28
[WIP] Experimental support for converting tile columns into a tensor

metasim:


metasim:
@wesm Thanks much for the follow-up. Will likely take you up on the offer as we revisit this in the next couple weeks. One of the things I'm concerned about is that even if we implement it using proper Arrow constructs, Spark's use of it may not yet accommodate Extension Types (hope I'm wrong):

https://github.com/apache/spark/blob/2856398de9d35d136758bbc11afa4d1dc0c98830/sql/catalyst/src/main/scala/org/apache/spark/sql/util/ArrowUtils.scala#L36-L57

wesm:
You might open a Spark JIRA issue about what you need there so it's tracked at least

metasim:
Possibly relevant: https://issues.apache.org/jira/browse/SPARK-27396

PR #342, opened 2019-09-11
Zonal stats page and other refactoring

### LINKED ISSUE #255: Vector Data Page Visual
Good spot for a visual would be at the end of the `Vector Data` page, which filters Landsat extents within 500 km of a selected point. Could show the geometries overlaid on the Landsat image.

### LINKED ISSUE #302: Folium map embedding improvments
The current state of the art for embedding a folium map in the documents could use some improvement: 

https://github.com/locationtech/rasterframes/blob/f91193b874db8d8ca4fe53a249e354a8251d1b26/pyrasterframes/src/main/python/docs/time-series.pymd#L45-L61

The results also could use some styling help. The map is  not tall enough. https://1258-174014757-gh.circle-artifacts.com/0/rf-site/time-series.html#vector-data

### LINKED ISSUE #340: Add section on zonal statistics
This could serve as a good start:

https://gist.github.com/metasim/76d11b89ae14524ceb75f16ce5e6973d

vpipkt:
1. Add a page in raster processing called zonal map algebra to fix #340 
2. Remove much discussion from time series page and refer to zonal page
3. Reference zonal page on the aggregations and vector data pages
4. Fix #302 
5. Fix #255


pyrasterframes/src/main/python/docs/__init__.py
```diff
@@ -40,6 +40,31 @@ def resource_dir_uri():
     return 'file://' + resource_dir()
 
 
+def folium_map(vector_data):
```
metasim: @vpipkt The one problem with this is that I'd like to get rid of `docs/__init__.py` So that the notebook versions can be self-contained. I can't remember the original reason this is required (as I thought `folium` has `display` methods)... Is it just for the `html` version?

pyrasterframes/src/main/python/docs/raster-processing.md
```diff
@@ -2,8 +2,9 @@
 
 @@@ index
 
-* @ref:[Local Algebra](local-algebra.md)
+* @ref:[Local Map Algebra](local-algebra.md)
 * @ref:["NoData" Handling](nodata-handling.md)
```
metasim: Should we move this so the "algebras" are near each other?

pyrasterframes/src/main/python/docs/time-series.pymd
```diff
@@ -5,6 +5,8 @@
 ```python setup, echo=False
 from IPython.display import display
 
+from docs import folium_map
```
metasim: Is this something we could perhaps handle in `rf_ipython` through some monkey patching?

pyrasterframes/src/main/python/docs/__init__.py
vpipkt: good thought there.

pyrasterframes/src/main/python/docs/raster-processing.md
vpipkt: my thinking was that the zonal leans on the same ideas as masking.

But we could make it stand alone more and reordre.

pyrasterframes/src/main/python/docs/__init__.py
vpipkt: So the formatter for `text/markdown` was the key but it seems that needed to be done outside `docs/__init__.py` script.

pyrasterframes/src/main/python/docs/vector-data.pymd
```diff
@@ -92,8 +107,21 @@ l8 = l8.withColumn('paducah', st_point(lit(-88.628), lit(37.072)))  # col of poi
...
+    l8_filtered.select('geom', 'bounds_wgs84').toPandas(),
+    crs='EPSG:4326',
+    geometry='geom')
+
+# display as folium / leaflet map
+m = folium.Map()
+layer = folium.GeoJson(geo_df.to_json())
+m.fit_bounds(layer.get_bounds())
+m.add_child(layer)
+m
```
vpipkt: This now renders correctly.

pyrasterframes/src/main/python/pyrasterframes/rf_ipython.py
```diff
@@ -178,9 +191,17 @@ def spark_df_to_html(df, num_rows=5, truncate=False):
         html_formatter.for_type(pandas.DataFrame, pandas_df_to_html)
         html_formatter.for_type(pyspark.sql.DataFrame, spark_df_to_html)
 
+        # these will likely only effect docs build
         markdown_formatter = ip.display_formatter.formatters['text/markdown']
         markdown_formatter.for_type(pyspark.sql.DataFrame, spark_df_to_markdown)
 
+        try:
```
vpipkt: @metasim thoughts about this?

metasim: I think this is OK, at least for now.

metasim: Definitely an improvement!!

metasim:
:shipit:

pyrasterframes/src/main/python/docs/__init__.py
metasim: @vpipkt 

> So the formatter for text/markdown was the key but it seems that needed to be done outside docs/__init__.py script.

Not sure I know what you mean... Do you have reservations about `pyrasterframes/rf_ipython.py`?

PR #345, opened 2019-09-13
Fixed Exception swallowing in RasterSource.

### LINKED ISSUE #267: Python access to RasterSource doesn't expose JVM errors
See #263  and #266 

Trying to take actions on these dataframes may result in a JVM stack trace with ERROR message that does not get to the JVM user. PYthon user just gets for example 0 `count` or empty list from a `take`

metasim:
Depends on #344 .

Fixes #267

### DIFF
--- a/core/src/main/scala/org/locationtech/rasterframes/expressions/generators/RasterSourceToRasterRefs.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/expressions/generators/RasterSourceToRasterRefs.scala
@@ -21,7 +21,6 @@
 
 package org.locationtech.rasterframes.expressions.generators
 
-import com.typesafe.scalalogging.LazyLogging
 import geotrellis.raster.GridBounds
 import geotrellis.vector.Extent
 import org.apache.spark.sql.catalyst.InternalRow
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/locationtech/rasterframes/expressions/generators/RasterSourceToRasterRefs.scala
```diff
@@ -77,9 +76,9 @@ case class RasterSourceToRasterRefs(children: Seq[Expression], bandIndexes: Seq[
     }
     catch {
       case NonFatal(ex) ⇒
-        val payload = Try(children.map(c => RasterSourceType.deserialize(c.eval(input)))).toOption.toSeq.flatten
-        logger.error("Error fetching data for one of: " + payload.mkString(", "), ex)
-        Traversable.empty
+        val description = Try(children.map(c => RasterSourceType.deserialize(c.eval(input))))
+          .toOption.toSeq.flatten.mkString(", ")
```
vpipkt: Hmm we have lost the pretty useful text of "Error fetching data for one of" 

Recommend reinstating that.

metasim: Ooops. Not intended. Good catch.

PR #359, opened 2019-09-19
WIP Fix Windows local GDAL read 

vpipkt:
Ensure our modifications to the incoming URI string work downstream in `VSIPath` for Windows local files.


vpipkt:
See @MiguelNOX contributed [output here](https://pastebin.com/6aaRXHW7)

Punch line is this: `Caused by: java.net.URISyntaxException: Illegal character in path at index 8: file:/D:\TFM\ACF\ACF_NDVI_savgolfix.envi` ... working on the pattern for windows reading here obv but getting a little closer.

PR #373, opened 2019-09-30
Fixed eager loading of RasterRef fields.

### LINKED ISSUE #372: Cannot fully delay header read when using RasterRefTile
The point of `RasterRef`s is to delay the call to a `RasterSource` until it's absolutely needed, including the encoding into Catalyst memory. Currently the serialization process asks for the `CRS`, `Extent` and `TileDimensions`, due to how tiles encode refs down at the cells level.

metasim:
Closes #372 .

### DIFF
--- a/core/src/main/scala/org/locationtech/rasterframes/ref/DelegatingRasterSource.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/ref/DelegatingRasterSource.scala
@@ -43,7 +43,7 @@ abstract class DelegatingRasterSource(source: URI, delegateBuilder: () => GTRast
       f(_delRef)
     }
     catch {
-      // On this exeception we attempt to recreate the delegate and read again.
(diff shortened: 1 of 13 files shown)


core/src/main/scala/org/locationtech/rasterframes/ref/DelegatingRasterSource.scala
```diff
@@ -43,7 +43,7 @@ abstract class DelegatingRasterSource(source: URI, delegateBuilder: () => GTRast
       f(_delRef)
     }
     catch {
-      // On this exeception we attempt to recreate the delegate and read again.
+      // On this Exeception we attempt to recreate the delegate and read again.
```
vpipkt: 😈 still mis-spelled `Exception`

metasim: 🤦‍♂

PR #375, opened 2019-10-03
Fixes docs build error due to missing library in Python build configuration

metasim:
Test build:

https://circleci.com/gh/s22s/rasterframes/1565


pyrasterframes/src/main/python/setup.py
```diff
@@ -140,6 +140,7 @@ def dest_file(self, src_file):
     author_email='info@astraea.earth',
     license='Apache 2',
     url='https://rasterframes.io',
+    python_requires=">=3.7",
```
vpipkt: I would guess that might be too strict. On my laptop my default is python 3.6. I am not sure how far back to allow.

pyrasterframes/src/main/python/setup.py
```diff
@@ -170,7 +171,7 @@ def dest_file(self, src_file):
         'folium',
     ],
     tests_require=[
-        'pytest==3.4.2',
+        'pytest>4.0.0,<5.0.0',
```
vpipkt: this seems beneficial.

PR #381, opened 2019-10-07
Databricks DBR 5.5 LTS compatibility

metasim:
Main benefit of this is that it provides Spark 2.4 compatibility.


pyrasterframes/src/main/python/scene_30_27_model.ipynb
```diff
@@ -0,0 +1,655 @@
+{
```
vpipkt: Looks like this got committed inadvertently ... please remove.

pyrasterframes/src/main/python/setup.py
```diff
@@ -130,6 +130,26 @@ def initialize_options(self):
...
+requests = 'requests'
+pytest_runner = 'pytest-runner'
+setuptools = 'setuptools>=0.8'
+ipython = 'ipython==6.2.1'
+ipykernel = 'ipykernel==4.8.0'
+pweave = 'Pweave==0.30.3'
+fiona = 'fiona==1.8.6'
+rasterio = 'rasterio>=1.0.0'
+folium = 'folium'
+pytest = 'pytest>4.0.0,<5.0.0'
```
vpipkt: possibly pytest>=4.0.0 ?

metasim:
Build associated with `40d8a35` failed only due to some time limit (it was otherwise running fine):

```console
The job exceeded the maximum time limit for jobs, and has been terminated.
```

vpipkt:
Build output looks good but it is saying it failed. I don't know what that is about. I would say :shipit:

Issue #383, opened 2019-10-10
in rf_ipython custom format for geometry objects


vpipkt:
To accomodate tiles we have unset the pandas max col width. 

However this commonly creates an unsatisfying situation where a geometry column is rendered as hundreds to thousands of charadcters of WKT.

We should try to create a custom rendering for geometries to limit these to whatever the user's original limit was? Can we do this for other objects (str?)  as well or is that a can of worms?

PR #385, opened 2019-10-11
In rf_ipython, honor users pandas max_colwidth on Geometry objects

vpipkt:
Refactors notebooks to use `pyrasterframes.utils.create_rf_spark_session` as well as update outputs with changes in `rf_ipython`


vpipkt:
Closes #383

Issue #310, opened 2019-08-23
Convert reference.pymd into a non-evaluated documentation file


metasim:
Eventually we'll want all these docs ported to the source and generated via Sphinx, etc. Not having to evaluate them might help with this process?

> not sure we have anything that runs any more? possibly this file can be converted to `md` instead of `pymd`

_Originally posted by @vpipkt in https://github.com/locationtech/rasterframes/pull/301_

PR #388, opened 2019-10-16
Add rf_local_is_in function

vpipkt:
To dos:

* tests currently failing. Some kind of type error in encoding or something
* add function to docs
* add Python bindings
* update masking docs to use it.


core/src/test/scala/org/locationtech/rasterframes/RasterFunctionsSpec.scala
```diff
@@ -972,4 +972,25 @@ class RasterFunctionsSpec extends TestEnvironment with RasterMatchers {
     val dResult = df.select($"ld").as[Tile].first()
     dResult should be (randNDPRT.localDefined())
   }
+
+  it("should check values isin"){
+    checkDocs("rf_local_is_in")
+
+    // tile is 3 by 3 with values, 1 to 9
+    val df = Seq((byteArrayTile, lit(1), lit(5), lit(10))).toDF("t", "one", "five", "ten")
```
metasim: @vpipkt Change `Seq((byteArrayTile, lit(1), lit(5), lit(10)))` to `Seq((byteArrayTile, 1, 5, 10))` and the `java.lang.UnsupportedOperationException: No Encoder found for org.apache.spark.sql.Column` error should go away. You only want to use `lit`after the Dataframe is created.

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/IsIn.scala
```diff
@@ -0,0 +1,88 @@
...
+  usage = "_FUNC_(tile, rhs) - In each cell of `tile`, return true if the value is in rhs.",
+  arguments = """
+  Arguments:
+    * tile - tile column to apply abs
+    * rhs - array to test against
+  """,
+  examples = """
+  Examples:
+    > SELECT  _FUNC_(tile, array);
+       ..."""
```
metasim: I think this is a case where I'd be helped by an example... I'm pretty sure I understand the use case/behavior now, but, but it took a bit of brow furrowing.

metasim: Of course `reference.md` too, when you get the implementation done.

metasim: And `release-notes.md`

vpipkt:
I am building the docs locally just to make sure the fix of #310 doesn't break anything.

@metasim let me know if docs look good and resolve your questions. We can embed a very short example.

vpipkt:
Local docs build was successful.

vpipkt:
See #400 which triggers circle CI docs build so we can see the final docs.

PR #397, opened 2019-10-21
Applying pre-partitioning to DataSources.

metasim:
Speeds up `experimental/it:test` from 527s to 194s.


vpipkt:
so nice I approved it twice!

PR #400, opened 2019-10-23
Add rf_local_is_in function 

### LINKED ISSUE #310: Convert reference.pymd into a non-evaluated documentation file

### LINKED ISSUE #351: Confusing use of `rf_rgb_composite` in supervised learning docs
At the end of the supervised learning page we are trying to show a side by side of the predicted label vs RGB image.

Since [this change](https://github.com/locationtech/rasterframes/commit/9dc6a63eaae6e168efb6f4668e87fb65d71c34d7#diff-846fcc30de7613b5c05f5019b5cc76daR222) went in the RGB image is actually a (128, 128) array and matplotlib is styiling it with viridis.

This is an antipattern and a misuse of `rf_rgb_composite`, as i understand it.

We should re-think the way we do this visualization to make a correct RGB and possibly to enable the visual to cover an expanded geographic scope.

vpipkt:
Add function to return an array indicating if cell values are in a provided array or python list.

Update docs to show this in masking

Also resolving these:

Close #351 
Close #310


vpipkt:
Docs as rendered: https://1690-174014757-gh.circle-artifacts.com/0/rf-site/supervised-learning.html#visualize-prediction

PR #408, opened 2019-11-05
Added forced truncation of WKT types in Markdown/HTML rendering.

### LINKED ISSUE #407: rf_ipython custom format for geometry objects in spark DataFrame
Related to  #383, basically fix the same condition for the markdown and html renderers of Spark DataFrames. The fix for #383 only corrected pandas DF.

metasim:
Closes #407

### DIFF
--- a/core/src/main/scala/org/locationtech/rasterframes/util/DataFrameRenderers.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/util/DataFrameRenderers.scala
@@ -24,12 +24,14 @@ package org.locationtech.rasterframes.util
 import geotrellis.raster.render.ColorRamps
 import org.apache.spark.sql.Dataset
 import org.apache.spark.sql.functions.{base64, concat, concat_ws, length, lit, substring, when}
+import org.apache.spark.sql.jts.JTSTypes
 import org.apache.spark.sql.types.{StringType, StructField}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/locationtech/rasterframes/util/DataFrameRenderers.scala
```diff
@@ -47,8 +49,9 @@ trait DataFrameRenderers {
               lit("\"></img>")
```
metasim: Wondering if the line above this one
```
 base64(rf_render_png(rf_resample(resolved, 0.5), ColorRamps.Viridis))
```
should be changed to use `ColorRamps.greyscale(256)`, and then apply a color ramp in the python API for notebooks.

vpipkt: I guess I am not sure about what we can do on the notebook / python side if the image is already b64 encoded in HTML... Some css stuff? idk

vpipkt:
🚢 it

core/src/main/scala/org/locationtech/rasterframes/util/DataFrameRenderers.scala
metasim: Good point. Damn. Didn't think of that.

PR #429, opened 2019-11-21
Add rf_agg_approx_quantiles function

vpipkt:
Initial implementation with quantiles

### To Do

- [x] Add to python API
- [ ] Add to SQL API
- [x] Docs
- [x] release notes


core/src/test/scala/org/locationtech/rasterframes/RasterFramesStatsSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.spark.sql.functions.col
+
+import org.locationtech.rasterframes._
+import org.locationtech.rasterframes.RasterFunctions
+
+class RasterFramesStatsSpec extends TestEnvironment with TestData {
+
+  describe("DataFrame.tileStats extension methods") {
+
+    val df = TestData.sampleGeoTiff.toDF()
```
vpipkt: would like to see a test of this with projected raster tile, raster ref, raster source etc. maybe in integration tests?

core/src/main/scala/org/locationtech/rasterframes/extensions/DataFrameMethods.scala
```diff
@@ -155,6 +155,9 @@ trait DataFrameMethods[DF <: DataFrame] extends MethodExtensions[DF] with Metada
   def withPrefixedColumnNames(prefix: String): DF =
     self.columns.foldLeft(self)((df, c) ⇒ df.withColumnRenamed(c, s"$prefix$c").asInstanceOf[DF])
 
+  /**  */
+  def tileStat(): RasterFrameStatFunctions = new RasterFrameStatFunctions(self)
```
metasim: Working on an alternative approach to discuss, one that takes an approach more in line with the other aggregate functions (i.e. `rf_agg_foo_bar`). While I know that Spark DataFrames uses this extension method approach in a couple of places (`stats`, `na`, ...) I think it adds an inconsistency that can feel arbitrary if the same thing can be achieved with columnar functions. `na` is a good example of one that can't, because it's effectively implementing a builder for NA handling, but I don't think were in that situation.

core/src/main/scala/org/locationtech/rasterframes/extensions/RasterFrameStatFunctions.scala
```diff
@@ -0,0 +1,76 @@
...
+      *   If set to zero, the exact quantiles are computed, which could be very expensive.
+      *   Note that values greater than 1 are accepted but give the same result as 1.
+      * @return the approximate quantiles at the given probabilities
+      *
+      * @note null and NaN values will be removed from the numerical column before calculation. If
+      *   the dataframe is empty or the column only contains null or NaN, an empty array is returned.
+      *
+      * @since 2.0.0
+      */
+    def approxTileQuantile(
```
metasim: Is this copied? If so, we need to attribute and reference, etc.

core/src/test/scala/org/locationtech/rasterframes/RasterFramesStatsSpec.scala
metasim: We can get there with Expressions...

core/src/test/scala/org/locationtech/rasterframes/RasterFramesStatsSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+
+      // computing externally with numpy we arrive at 7963, 10068, 12160 for these quantiles
+      result should contain inOrderOnly (7963.0, 10068.0, 12160.0)
+    }
+
+    it("should compute approx percentiles for many tile cols"){
+      val result = df.tileStat().approxTileQuantile(
+        Array("tile", "tilePlus2"),
+        Array(0.25, 0.75),
+        0.00001
```
metasim: Is there any way to default this? Or is it strongly scale dependent?

metasim:
See this for a couple alternate approaches that don't depend DataFrame extension methods.

https://github.com/s22s/rasterframes/pull/57

core/src/main/scala/org/locationtech/rasterframes/extensions/RasterFrameStatFunctions.scala
vpipkt: ya oops

vpipkt:
@metasim tests failing on the SQL expression. Not sure how to work in all the different parameters. We could easily register some functions to compute default percentiles like median, quartiles, quantiles, deciles, etc at default relative error.

metasim:
@vpipkt  I think enabling SQL support is going to take some more work, given the SQL function parameter requirements.... perhaps we need to look for something in the official API that uses non-columnar parameters to a function (does it exist)? For some reason I'm loath to make all of those parameters columnar, but perhaps that's the proper way to do it

I think the current question is do we merge this without support for SQL, or do we work this out first?

vpipkt:
I am fine with a little divergence in the API between SQL and python/scala over this. 

Possible paths:

   1. No SQL function. In docs state there is no SQL support.
   1. Expose SQL functions for "canned" quantiles; Can add these to Python and Scala as well
       1. `rf_agg_approx_median`
       1. `rf_agg_approx_quartiles`
       1. `rf_agg_approx_quantiles` 
       1. `rf_agg_approx_deciles`

vpipkt:
Latest commit should pass tests and is basically the option 1 [above](https://github.com/locationtech/rasterframes/pull/429#issuecomment-570672917)

Recommend we add an issue for option 2, lower priority.

core/src/main/scala/org/locationtech/rasterframes/expressions/aggregates/ApproxCellQuantilesAggregate.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  override def update(buffer: MutableAggregationBuffer, input: Row): Unit = {
+    val qs = buffer.getStruct(0).to[QuantileSummaries]
+    if (!input.isNullAt(0)) {
+      val tile = input.getAs[Tile](0)
+      var result = qs
+      tile.foreachDouble(d => if (!isNoData(d)) result = result.insert(d))
+      buffer.update(0, result.toRow)
+    }
+    else buffer
```
vpipkt: Compiler is complaining about this. Since the function returns `Unit` this line could be omitted?

metasim: Yeh, that was me not thinking "mutably". It can go.

PR #430, opened 2019-11-21
Aggregate overview raster

metasim:

PR #433, opened 2019-11-22
Fix for 409: assert on mask target must have NoData defined

### LINKED ISSUE #409: masking functions should warn or convert if raw cell type
Masking seems "broken" when the data tile that the mask is applied to has no nodata defined (raw celltype)

Queries about nodata cell counts are inconsistent with tile contents of the mask .

The masking functions could:

1. throw assertion error, fail outright if the celltype of the data tile has no nodata;
1. warn the user in logs but continue with the current logic;
1. implicitly convert the raw type to the default nodata value; __OR__
1. rely on a user config to decide among above options

Failing has the downside that it may not be evaluated until other parts of the spark job graph have been worked. REsulting in wasted computation and frustration. 

Warnings may not be seen or noticed by the user ... esp true for users in the python api

Converting the celltype may mark valid data as nodata without the user understanding quite why, impacting the result of analytics.

Even with a user config there is the issue of setting a reasonable default.

vpipkt:
Fixes #409

### DIFF
--- a/core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/Mask.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/Mask.scala
@@ -23,7 +23,7 @@ package org.locationtech.rasterframes.expressions.transformers
 
 import com.typesafe.scalalogging.Logger
 import geotrellis.raster
-import geotrellis.raster.{CellType, NoNoData, Tile}
+import geotrellis.raster.{NoNoData, Tile}
 import geotrellis.raster.mapalgebra.local.{Undefined, InverseMask ⇒ gtInverseMask, Mask ⇒ gtMask}
 import org.apache.spark.sql.catalyst.analysis.TypeCheckResult
 import org.apache.spark.sql.catalyst.analysis.TypeCheckResult.{TypeCheckFailure, TypeCheckSuccess}
@@ -74,13 +74,8 @@ abstract class Mask(val left: Expression, val middle: Expression, val right: Exp
     implicit val tileSer = TileUDT.tileSerializer
     val (targetTile, targetCtx) = tileExtractor(targetExp.dataType)(row(targetInput))
 
-    // Which of these is preferable? companion object Seq contains or pattern match on trait?
-//    assert(!CellType.noNoDataCellTypes.contains(targetTile.cellType), maskErrorStr)
-    targetTile.cellType match {
-      case _: NoNoData ⇒ assert(false,
-        s"Input data expression ${left.prettyName} must have a CellType with NoData defined in order to perform a masking operation. Found CellType ${targetTile.cellType.toString()}.")
-      case _ ⇒
-    }
+    require(! targetTile.cellType.isInstanceOf[NoNoData],
+      s"Input data expression ${left.prettyName} must have a CellType with NoData defined in order to perform a masking operation. Found CellType ${targetTile.cellType.toString()}.")
 
     val (maskTile, maskCtx) = tileExtractor(maskExp.dataType)(row(maskInput))
 
--- a/core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/Mask.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/Mask.scala
@@ -23,7 +23,7 @@ package org.locationtech.rasterframes.expressions.transformers
 
 import com.typesafe.scalalogging.Logger
 import geotrellis.raster
-import geotrellis.raster.Tile
+import geotrellis.raster.{CellType, NoNoData, Tile}
 import geotrellis.raster.mapalgebra.local.{Undefined, InverseMask ⇒ gtInverseMask, Mask ⇒ gtMask}
 import org.apache.spark.sql.catalyst.analysis.TypeCheckResult
 import org.apache.spark.sql.catalyst.analysis.TypeCheckResult.{TypeCheckFailure, TypeCheckSuccess}
@@ -73,6 +73,15 @@ abstract class Mask(val left: Expression, val middle: Expression, val right: Exp
   override protected def nullSafeEval(targetInput: Any, maskInput: Any, maskValueInput: Any): Any = {
     implicit val tileSer = TileUDT.tileSerializer
     val (targetTile, targetCtx) = tileExtractor(targetExp.dataType)(row(targetInput))
+
+    // Which of these is preferable? companion object Seq contains or pattern match on trait?
+//    assert(!CellType.noNoDataCellTypes.contains(targetTile.cellType), maskErrorStr)
+    targetTile.cellType match {
+      case _: NoNoData ⇒ assert(false,
+        s"Input data expression ${left.prettyName} must have a CellType with NoData defined in order to perform a masking operation. Found CellType ${targetTile.cellType.toString()}.")
+      case _ ⇒
+    }
+
     val (maskTile, maskCtx) = tileExtractor(maskExp.dataType)(row(maskInput))
 
     if (targetCtx.isEmpty && maskCtx.isDefined)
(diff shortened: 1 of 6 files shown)


core/src/test/scala/org/locationtech/rasterframes/MaskingFunctionsSpec.scala
```diff
@@ -0,0 +1,439 @@
...
+import org.apache.spark.sql.Encoders
+import org.apache.spark.sql.functions._
+import org.locationtech.rasterframes.expressions.accessors.ExtractTile
+import org.locationtech.rasterframes.stats._
+import org.locationtech.rasterframes.tiles.ProjectedRasterTile
+
+
+import geotrellis.raster.testkit.RasterMatchers
+
+class MaskingFunctionsSpec extends TestEnvironment with RasterMatchers {
```
vpipkt: Note we have refactored the mask related function tests out into this spec. Just because we had RasterFunctionsSpec gettting so big and starting to get disorganized.

core/src/test/scala/org/locationtech/rasterframes/MaskingFunctionsSpec.scala
```diff
@@ -0,0 +1,439 @@
...
+
+      val cellTyperesult2 = CellType.noNoDataCellTypes.contains(noNoDataCellType)
+      cellTyperesult2 should be (true)
+
+      val df = Seq(TestData.projectedRasterTile(5, 5, 42, TestData.extent, TestData.crs,
+        noNoDataCellType))
+        .toDF("tile")
+      // if this had NoData defined would be a no-op because the tile is all datacells
+      lazy val result = df.select(rf_mask($"tile", $"tile")).collect()
+      an[AssertionError] should be thrownBy(result)
```
vpipkt: The meat of this PR.

core/src/test/scala/org/locationtech/rasterframes/MaskingFunctionsSpec.scala
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    it("should throw if no nodata"){
+      val noNoDataCellType = UByteCellType
+      val cellTypeResult = noNoDataCellType match {
+        case _: NoNoData ⇒ 99
+        case _ ⇒ 0
+      }
+//      cellTypeResult should begreater (0)
+      cellTypeResult should be >(0)
```
vpipkt: these several lines are cruft.

core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/Mask.scala
```diff
@@ -73,6 +73,15 @@ abstract class Mask(val left: Expression, val middle: Expression, val right: Exp
   override protected def nullSafeEval(targetInput: Any, maskInput: Any, maskValueInput: Any): Any = {
     implicit val tileSer = TileUDT.tileSerializer
     val (targetTile, targetCtx) = tileExtractor(targetExp.dataType)(row(targetInput))
+
+    // Which of these is preferable? companion object Seq contains or pattern match on trait?
```
vpipkt: @metasim what do you prefer here?

metasim: Definitely not `contains`... too wordy. You could also do `targetTile.cellType.isInstanceOf[NoNoData]`.

core/src/test/scala/org/locationtech/rasterframes/MaskingFunctionsSpec.scala
metasim: Put this in the `functions` package. I've continued down this vein in https://github.com/locationtech/rasterframes/pull/430

core/src/test/scala/org/locationtech/rasterframes/MaskingFunctionsSpec.scala
metasim: I prefer:

```scala
an[AssertionError] should be thrownBy {
  df.select(rf_mask($"tile", $"tile")).collect()
}
```

core/src/test/scala/org/locationtech/rasterframes/MaskingFunctionsSpec.scala
```diff
@@ -0,0 +1,430 @@
...
+        .withColumn("qa_fill", rf_local_extract_bits($"mask", lit(0)))
+        .withColumn("qa_sat", rf_local_extract_bits($"mask", lit(2), lit(2)))
+        .withColumn("qa_cloud", rf_local_extract_bits($"mask", lit(4)))
+        .withColumn("qa_cconf", rf_local_extract_bits($"mask", 5, 2))
+        .withColumn("qa_snow", rf_local_extract_bits($"mask", lit(9), lit(2)))
+        .withColumn("qa_circonf", rf_local_extract_bits($"mask", 11, 2))
+
+      def checker(colName: String, valFilter: Int, assertValue: Int): Unit = {
+        // print this so we can see what's happening if something  wrong
+        println(s"${colName} should be ${assertValue} for qa val ${valFilter}")
```
metasim: Send to `logger.debug` instead.

core/src/test/scala/org/locationtech/rasterframes/MaskingFunctionsSpec.scala
```diff
@@ -0,0 +1,430 @@
...
+        /**  in this unit test setup, the `val` column is an integer that the entire row's mask is full of
+          * filter for the maskValueFilter
+          * then check the columnName and look at the masked data tile given by `columnName`
+          * assert that the `columnName` tile is / is not all nodata based on `resultIsNoData`
+          * */
+
+        val printOutcome = if (resultIsNoData) "all NoData cells"
+        else "all data cells"
+
+        println(s"${columnName} should contain ${printOutcome} for qa val ${maskValueFilter}")
```
metasim: `logger.debug`

core/src/test/scala/org/locationtech/rasterframes/MaskingFunctionsSpec.scala
```diff
@@ -0,0 +1,430 @@
...
+        val resultDf = result
+          .filter($"val" === lit(maskValueFilter))
+
+        val resultToCheck: Boolean = resultDf
+          .select(rf_is_no_data_tile(col(columnName)))
+          .first()
+
+        val dataTile = resultDf.select(col(columnName)).as[ProjectedRasterTile].first()
+        println(s"\tData tile values for col ${columnName}: ${dataTile.toArray().mkString(",")}")
```
metasim: `logger.debug`

docs/src/main/paradox/release-notes.md
```diff
@@ -8,6 +8,7 @@
 * _Breaking_: `rf_spatial_index` renamed `rf_xz2_index` to differentiate between XZ2 and Z2 variants.
 * Added `withSpatialIndex` to RasterSourceDataSource to pre-partition tiles based on tile extents mapped to a Z2 space-filling curve 
 * Add `rf_mask_by_bit`, `rf_mask_by_bits` and `rf_local_extract_bits` to deal with bit packed quality masks. Updated the masking documentation to demonstrate the use of these functions.
+* Throw an `AssertionError` when attempting to apply a mask to a `Tile` whose `CellType` has no NoData defined. ([#409](https://github.com/locationtech/rasterframes/issues/384))
```
metasim: I'm questioning whether `AssertionError` is the right one... `require` uses `IllegalArgumentException`, so I think this is more in line with conventional semantics.

core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/Mask.scala
```diff
@@ -73,6 +73,15 @@ abstract class Mask(val left: Expression, val middle: Expression, val right: Exp
...
+//    assert(!CellType.noNoDataCellTypes.contains(targetTile.cellType), maskErrorStr)
+    targetTile.cellType match {
+      case _: NoNoData ⇒ assert(false,
```
metasim: See comment below about `AssertionError`. The convention I see in Scala is that `require` is more common for user-sourced errors, and `assert` is for trapping coder-sourced errors. So I think `assert` should be changed to `require`.

docs/src/main/paradox/release-notes.md
vpipkt: I think you are right.

PR #434, opened 2019-12-05
TensorUDT prototype

echeipesh:
Prototype to use Arrow Tensor IPC message to add a tensor UDT.

The aim are twofold:
- Add a tensor UDT which is the main objective for our target workflows
- Use some kind of arrow encoding in hopes that it can be improved on

Its worth noting that in current world the bytes will still passed through python pickler on the way to python executors, which can't be great for performance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ tour/*.tiff
 scoverage-report*
 
 zz-*
+venv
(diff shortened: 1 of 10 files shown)


echeipesh:
Sorry for PR spam here, prototype PR against our fork, not worth attention ATM.

PR #435, opened 2019-12-05
Add libspatialindex and rtree to test environment

vpipkt:
Enables use of geopandas spatial join in the docs and other parts of the build process, and for downstream apps.


vpipkt:
> Just to confirm: you pushed the `spatialIndex` tag to Docker Hub?

No. I did not.

PR #446, opened 2020-01-09
Docs/fix 252 viz extents

### LINKED ISSUE #252: Visual for scene vs tile extents

vpipkt:
Fix #252 

Stacks on top of #430

Add a folium map to the `raster-read` page showing tile overview and extents.

image


vpipkt:
Lookin good now

Issue #466, opened 2020-02-14
GeoTrellisRelation: unsafe symbol Logger in runtime reflection universe


jdenisgiguere:
## Current situation

When I attempt to access a Geotrellis Layer with `pyrasterframes`, I get the error given in the log below.

```python
from pyrasterframes.utils import create_rf_spark_session
if __name__ == '__main__':
    spark = create_rf_spark_session()
    catalog = spark.read.geotrellis_catalog("s3://my-bucket/gt3catalog")
    catalog.printSchema()
    layer = spark.read.format("geotrellis").option("layer", "my_layer").option("zoom", 0).load("s3://my-bucket/gt3catalog").as_layer()
    layer.printSchema() # OK
    layer.select("geometry").collect() #  Oups...
```

In interactive mode, I will encounter the error once. The next invocation, it will be successful.

This seems related to #293, which is fixed. cc @metasim.

<details>

```
py4j.protocol.Py4JJavaError: An error occurred while calling o72.collectToPython.
: java.lang.RuntimeException: error reading Scala signature of org.locationtech.rasterframes.datasource.geotrellis.GeoTrellisRelation: assertion failed: unsafe symbol Logger (child of <none>) in runtime reflection universe
	at scala.reflect.internal.pickling.UnPickler.unpickle(UnPickler.scala:46)
	at scala.reflect.runtime.JavaMirrors$JavaMirror.unpickleClass(JavaMirrors.scala:619)
	at scala.reflect.runtime.SymbolLoaders$TopClassCompleter$$anonfun$complete$1.apply$mcV$sp(SymbolLoaders.scala:28)
	at scala.reflect.runtime.SymbolLoaders$TopClassCompleter$$anonfun$complete$1.apply(SymbolLoaders.scala:25)
	at scala.reflect.runtime.SymbolLoaders$TopClassCompleter$$anonfun$complete$1.apply(SymbolLoaders.scala:25)
	at scala.reflect.internal.SymbolTable.slowButSafeEnteringPhaseNotLaterThan(SymbolTable.scala:263)
	at scala.reflect.runtime.SymbolLoaders$TopClassCompleter.complete(SymbolLoaders.scala:25)
	at scala.reflect.internal.Symbols$Symbol.info(Symbols.scala:1535)
[...]

PR #469, opened 2020-02-14
Unit tests: write a rasteframes to geotrellis with pyrasterframes

jdenisgiguere:
A test with a single band geotiff and a test with a multiband geotiff. The second is currently failing.

This is a step to reproduce #466, but it is not reproducible yet.

Signed-off-by: Jean-Denis Giguère <jean-denis@anagraph.io>


metasim:
@jdenisgiguere Do you think this is ready to merge?

jdenisgiguere:
@metasim, yes! This fix #466.

PR #488, opened 2020-05-14

vpipkt:
In local circleci command was able to get past the current job failure. Waiting for circle build to see if that's all we need.


vpipkt:
hmm think this branch name may not trigger the build.. :-/

PR #489, opened 2020-05-14

vpipkt:
In local circleci command was able to get past the current job failure. Waiting for circle build to see if that's all we need.


vpipkt:
Oops need to push to s22s fork to trigger CI build. 

It's been a minute ....

PR #490, opened 2020-05-14

vpipkt:
In local circleci command, I was able to get past the current job failure. Waiting for circle build to see if that's all we need.


vpipkt:
Now a slightly more substantive error. Looks like the JVM died.

https://app.circleci.com/pipelines/github/s22s/rasterframes/1207/workflows/41fd2968-aea9-4e5c-afac-0b76604f905c/jobs/2679/steps

PR #495, opened 2020-07-08
Choose ResampleMethod in rf_resample and RasterJoin

vpipkt:
Allows using a string argument in `rf_resample` and `RasterJoin` to choose the resampling method to use.

The choices are (either snake_case or CamelCase would be acceptable):

    nearest_neighbor, bilinear, cubic_convolution, cubic_spline, lanczos, average, mode, median, max, min, sum

See [GeoTrellis docs](https://geotrellis.github.io/scaladocs/2.0/#geotrellis.raster.resample.package) for more. 

To Do

- [x] Unit test for raster_join
- [x] Fix possible serialization issue in failing tests for `DataFrame.toMarkdown`
- [x] UPdate release notes
- [x] Update docs chapter about raster join
- [x] update function reference
- [x] ...?


core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
```diff
@@ -109,6 +110,7 @@ package object expressions {
     registry.registerExpression[ExpM1]("rf_expm1")
     registry.registerExpression[Sqrt]("rf_sqrt")
     registry.registerExpression[Resample]("rf_resample")
```
vpipkt, line 112: this is actually a breaking change. SQL code that passed only two arguments before will break.

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,171 @@
...
-  override protected def op(left: Tile, right: Tile): Tile = left.resample(right.cols, right.rows, NearestNeighbor)
-  override protected def op(left: Tile, right: Double): Tile = left.resample((left.cols * right).toInt,
-                                                                             (left.rows * right).toInt, NearestNeighbor)
-  override protected def op(left: Tile, right: Int): Tile = op(left, right.toDouble)
+  override def dataType: DataType = left.dataType
+  override def children: Seq[Expression] = Seq(left, right, method)
+
+  def targetFloatIfNeeded(t: Tile, method: ResampleMethod): Tile =
+     method match {
+      case NearestNeighbor | Mode | RMax | RMin | Sum  ⇒ t
```
vpipkt, line 54: would like to discuss this in the docs. basically these methods are guaranteed to return an integer type. However you can still get (with sum) some bad behavior like uint8 sum over a large neighborhood resulting in > 255.

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,171 @@
...
+      case _ ⇒ fpTile(t)
```
vpipkt, line 55: all other methods can return float value even if input tile is integral type

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,171 @@
...
       val r = right.eval(input)
-      if (l == null && r == null) null
-      else if (l == null) r
-      else if (r == null && tileExtractor.isDefinedAt(right.dataType)) l
-      else if (r == null) null
-      else nullSafeEval(l, r)
+      val m = method.eval(input)
+      if (m == null) null // no method, return null
+      else if (l == null) null  // no l tile, return null
+      else if (r == null) l // no target tile or factor, return l without changin it
```
vpipkt, line 114: Not sure but pretty sure this is sensible. says the semantics are:

```
rf_resample(t: Tile, <null>: Double, m: String) => t
rf_resample(t: Tile, <null>: Tile, m: String) => t 
```

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,171 @@
...
+    extends ResampleBase(tile, target, Literal("nearest"))
+  object ResampleNearest {
+    def apply(tile: Column, target: Column): Column =
+      new Column(ResampleNearest(tile.expr, target.expr))
+
+    def apply[N: Numeric](tile: Column, value: N): Column =
+      new Column(ResampleNearest(tile.expr, lit(value).expr))
+  }
+
+  def stringToMethod(methodName: String): ResampleMethod = {
```
vpipkt, line 176: this appears again. should refactor someplace

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,169 @@
...
+  override def dataType: DataType = left.dataType
+  override def children: Seq[Expression] = Seq(left, right, method)
+
+  def targetFloatIfNeeded(t: Tile, method: ResampleMethod): Tile =
+     method match {
+      case NearestNeighbor | Mode | RMax | RMin | Sum  ⇒ t
+      case _ ⇒ fpTile(t)
+     }
+
+  def stringToMethod(methodName: String): ResampleMethod =
```
vpipkt, line 57: refactor this to some common place. see RasterJoin code https://github.com/locationtech/rasterframes/pull/495/files#diff-9149ea4528867e501de8472e5596d256R112

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
metasim: Oooh, good points. Shoot.  Wish we could detect overflow. Does GeoTrellis have any comments on this?

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,168 @@
...
+    methodName.toLowerCase().trim().replaceAll("_", "") match {
```
metasim, line 57: :win:

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,168 @@
...
+
+   // These methods define the core algorithms to be used.
+  def op(left: Tile, right: Tile, method: String): Tile = {
+    val m = stringToMethod(method)
+    targetFloatIfNeeded(left, m)
+      .resample(right.cols, right.rows, m)
+  }
+
+  def op(left: Tile, right: Double, method: String): Tile = {
+    val m = stringToMethod(method)
```
metasim, line 80: Ideally if we had more time we'd profile to see if calling `stringToMethod` is negligible or not.

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,168 @@
...
+      case "cubicspline" ⇒ CubicSpline
+      case "lanczos" | "lanzos" ⇒ Lanczos
+      // aggregates
+      case "average" ⇒ Average
+      case "mode" ⇒ Mode
+      case "median" ⇒ Median
+      case "max" ⇒ RMax
+      case "min" ⇒ RMin
+      case "sum" ⇒ Sum
+    }
```
metasim, line 70: I think this code would do well to be put in `utils` or in the package object. The Scala-ic way would be as an extractor. Something like this (flying blind here):

```
object SMethod {
  def unapply(name: String): Option[ResamplingMethod] = {
     name.toLowerCase().trim().replaceAll("_", "") match {
      case "nearestneighbor" | "nearest" ⇒ Some(NearestNeighbor)
      case "bilinear" ⇒ Some(Bilinear)
      case "cubicconvolution" ⇒ Some(CubicConvolution)
      case "cubicspline" ⇒ Some(CubicSpline)
      case "lanczos" | "lanzos" ⇒ Some(Lanczos)
      // aggregates
      case "average" ⇒ Some(Average)
      case "mode" ⇒ Some(Mode)
      case "median" ⇒ Some(Median)
      case "max" ⇒ Some(RMax)
      case "min" ⇒ Some(RMin)
      case "sum" ⇒ Some(Sum)
      case _ => None
    }
  }
}
```

You could then us it like this:

```
m = name match {
  case SMethod(mth) => mth
  case _ => throw Barf("unrecognized")
}
```

or

```
SMethod(m) = name
```

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
metasim: Probably me being fussy, but occasionally I see in the profiler a lot of time spent in Scala collections or whatnot... usually a sign that a loop needs to be rethought.

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,168 @@
...
+  def op(left: Tile, right: Tile, method: String): Tile = {
+    val m = stringToMethod(method)
+    targetFloatIfNeeded(left, m)
+      .resample(right.cols, right.rows, m)
+  }
+
+  def op(left: Tile, right: Double, method: String): Tile = {
+    val m = stringToMethod(method)
+    targetFloatIfNeeded(left, m)
+      .resample((left.cols * right).toInt, (left.rows * right).toInt, m)
```
metasim, line 82: For documentation sake, can you introduce an intermediate variable for `left.cols * right` and `left.rows * right`? Something like `newRows` and `newCols` (if I'm capturing the intent properly).

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,168 @@
...
+
+    val (leftTile, leftCtx) = tileExtractor(left.dataType)(row(input1))
+    val methodString = input3.asInstanceOf[UTF8String].toString
+
+    val result: Tile = tileOrNumberExtractor(right.dataType)(input2) match {
+      // in this case we expect the left and right contexts to vary. no warnings raised.
+      case TileArg(rightTile, _) ⇒ op(leftTile, rightTile, methodString)
+      case DoubleArg(d) ⇒ op(leftTile, d, methodString)
+      case IntegerArg(i) ⇒ op(leftTile, i.toDouble, methodString)
+    }
```
metasim, line 110: I think I would argue for doing the `stringToMethod` in here and having the `op` methods take the real type. You've "reified" the other inputs already, seems like we should do it here. The issue we were discussing before relating to keeping it "stringly typed" until the last moment was a) partially due to me misreading the exception, and b) only applies to values making their way into the member fields of the class (where they'd get serialized).

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,168 @@
...
+
+    // reassemble the leftTile with its context. Note that this operation does not change Extent and CRS
```
metasim, line 112: Good point!

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,168 @@
...
+  * method_name - one the following options: nearest_neighbor, bilinear, cubic_convolution, cubic_spline, lanczos, average, mode, median, max, min, sum
+                  This option can be CamelCase as well
+""",
+  examples = """
+Examples:
+  > SELECT _FUNC_(tile, 0.2, median);
+     ...
+  > SELECT _FUNC_(tile1, tile2, lit("cubic_spline"));
+     ..."""
+)
```
metasim, line 150: Gosh, I'm so sick of having to have documentation in so many places...

metasim: Someday we'll figure out how to unify it.

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,168 @@
...
-)
-case class Resample(left: Expression, right: Expression) extends BinaryLocalRasterOp
-  with CodegenFallback {
   override val nodeName: String = "rf_resample"
-  override protected def op(left: Tile, right: Tile): Tile = left.resample(right.cols, right.rows, NearestNeighbor)
-  override protected def op(left: Tile, right: Double): Tile = left.resample((left.cols * right).toInt,
-                                                                             (left.rows * right).toInt, NearestNeighbor)
-  override protected def op(left: Tile, right: Int): Tile = op(left, right.toDouble)
+  override def dataType: DataType = left.dataType
+  override def children: Seq[Expression] = Seq(left, right, method)
```
metasim, line 48: 👍

core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
metasim: Thanks for documenting it! I think it's fine.

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
```diff
@@ -22,53 +22,168 @@
...
+    * tile - tile
+    * rhs  - scalar or tile to match dimension""",
+ examples = """
+  Examples:
+    > SELECT _FUNC_(tile, 2.0);
+       ...
+    > SELECT _FUNC_(tile1, tile2);
+       ...""")
+case class ResampleNearest(tile: Expression, target: Expression)
+  extends ResampleBase(tile, target, Literal("nearest"))
```
metasim, line 180: Since you want to override `nodeName` here to be `rf_resample_nearest`.

core/src/main/scala/org/locationtech/rasterframes/extensions/RasterJoin.scala
```diff
@@ -36,7 +36,7 @@ import scala.util.Random
 object RasterJoin {
```
metasim, line 36: I'm torn on what to do here with the string-ly typed resample methods. It's what I suggested as a hack around the  serialization error, but ultimately I'd want the Scala API to use first-class types as must as possible in the user-facing API. Even if that mean turning it into a string to instantiate the `Expression`. I don't want to add a bunch of onerous work, but I suspect I'll regret the public API not using strong type. We can leave it as is to get it out the door, then change it later as a breaking change in Scala (Python would remain string-ly), or make some extra effort here to get the Scala-land API ship shape. I'll let you decide.

core/src/main/scala/org/locationtech/rasterframes/extensions/ReprojectToLayer.scala
```diff
@@ -48,7 +48,7 @@ object ReprojectToLayer {
     // Create effectively a target RasterFrame, but with no tiles.
     val dest = gridItems.toSeq.toDF(SPATIAL_KEY_COLUMN.columnName, EXTENT_COLUMN.columnName, CRS_COLUMN.columnName)
 
-    val joined = RasterJoin(broadcast(dest), df, Some(tlm.tileLayout.tileDimensions))
+    val joined = RasterJoin(broadcast(dest), df, "nearest", Some(tlm.tileLayout.tileDimensions))
```
metasim, line 51: Why not expose the parameter here too? Can make that a default value to the `apply`.

core/src/main/scala/org/locationtech/rasterframes/functions/TileFunctions.scala
```diff
@@ -104,11 +104,21 @@ trait TileFunctions {
 
   /** Resample tile to different size based on scalar factor or tile whose dimension to match. Scalar less
    * than one will downsample tile; greater than one will upsample. Uses nearest-neighbor. */
-  def rf_resample[T: Numeric](tileCol: Column, factorValue: T) = Resample(tileCol, factorValue)
+  def rf_resample[T: Numeric](tileCol: Column, factorValue: T) = ResampleNearest(tileCol, factorValue)
```
metasim, line 107: In SQL this is called `rf_resample_nearest` (and as the `nodeName` per suggestion above). Is the rationale for backwards compatibility?

core/src/main/scala/org/locationtech/rasterframes/functions/TileFunctions.scala
```diff
@@ -104,11 +104,21 @@ trait TileFunctions {
...
+  def rf_resample(tileCol: Column, factorCol: Column) = ResampleNearest(tileCol, factorCol)
+
+  /**  */
+  def rf_resample[T: Numeric](tileCol: Column, factorVal: T, methodName: Column) = Resample(tileCol, factorVal, methodName)
+
+  def rf_resample[T: Numeric](tileCol: Column, factorVal: T, methodName: String) = Resample(tileCol, factorVal, methodName)
+
+  def rf_resample(tileCol: Column, factorCol: Column, methodName: Column) = Resample(tileCol, factorCol, methodName)
+
+  def rf_resample(tileCol: Column, factorCol: Column, methodName: String) = Resample(tileCol, factorCol, lit(methodName))
```
metasim, line 120: blah blah blah ;-)

docs/src/main/paradox/reference.md
```diff
@@ -154,12 +154,25 @@ Change the interpretation of the `tile_col`'s cell values according to specified
...
+    Tile rf_resample(Tile tile, Double factor, String method)
+    Tile rf_resample(Tile tile, Int factor, String method)
+    Tile rf_resample(Tile tile, Tile shape_tile, String method)
+    Tile rf_resample_nearest(Tile tile, Double factor)
+    Tile rf_resample_nearest(Tile tile, Int factor)
+    Tile rf_resample_nearest(Tile tile, Tile shape_tile)
+
+
+Change the tile dimension by upsampling or downsampling. Passing a numeric `factor` will scale the number of columns and rows in the tile: 1.0 is the same number of columns and row; less than one downsamples the tile; and greater than one upsamples the tile. Passing a tile as the second argument resamples such that the output has the same dimension (number of columns and rows) as `shape_tile`. Resampling methods can be one of: nearest_neighbor, bilinear, cubic_convolution, cubic_spline, lanczos, average, mode, median, max, min, or sum.
```
metasim, line 171: In the markdown, render the method names as `"resampling_method"`

Consider breaking them into two bullets, one list for aggregates, the other for non-.

docs/src/main/paradox/reference.md
```diff
@@ -154,12 +154,25 @@ Change the interpretation of the `tile_col`'s cell values according to specified
...
+
+Note the last six options apply aggregates when downsampling. For example a 0.25 factor and `max` method returns the maximum value in a 4x4 neighborhood.
```
metasim, line 173: Does the GT algorithm handle the edges as you'd expect?

docs/src/main/paradox/release-notes.md
```diff
@@ -2,6 +2,13 @@
 
 ## 0.9.x
 
+### 0.9.1
+
+* Added `method_name` parameter to the `rf_resample` method.
+   * __BREAKING__: In SQL, the function `rf_resample` now takes 3 arguments. You can use `rf_resample_nearest` with two arguments or refactor to `rf_resample(t, v, "nearest")`.
+* Added resample method parameter to SQL and Python APIs. This will affect the reprojection of right hand side tiles.
```
metasim, line 9: > This will affect the reprojection of right hand side tiles.
 
Can you be explicit in what way here?

pyrasterframes/src/main/python/docs/raster-join.pymd
```diff
@@ -17,7 +17,7 @@ A common operation for raster data is reprojecting or warping the data to a diff
   
 In RasterFrames, you can perform a **Raster Join** on two DataFrames containing raster data. 
 The operation will perform a _spatial join_ based on the [CRS][CRS] and [extent][extent] data in each DataFrame. By default it is a left join and uses an intersection operator.
-For each candidate row, all _tile_ columns on the right hand side are warped to match the left hand side's [CRS][CRS], [extent][extent], and dimensions. Warping relies on GeoTrellis library code and uses nearest neighbor resampling method. 
+For each candidate row, all _tile_ columns on the right hand side are warped to match the left hand side's [CRS][CRS], [extent][extent], and dimensions. Warping relies on GeoTrellis library code. You can specify the resampling method to be applied as one of: nearest_neighbor, bilinear, cubic_convolution, cubic_spline, lanczos, average, mode, median, max, min, or sum.
```
metasim, line 20: Same with markdown: `"resampling_method"`

pyrasterframes/src/main/python/pyrasterframes/__init__.py
```diff
@@ -72,21 +72,25 @@ def _convert_df(df: DataFrame, sp_key=None, metadata=None) -> RasterFrameLayer:
 def _raster_join(df: DataFrame, other: DataFrame,
                  left_extent=None, left_crs=None,
                  right_extent=None, right_crs=None,
-                 join_exprs=None) -> DataFrame:
+                 join_exprs=None, resampling_method='nearest_neighbor') -> DataFrame:
     ctx = SparkContext._active_spark_context._rf_context
+    assert resampling_method in ['nearest_neighbor', 'bilinear', 'cubic_convolution', 'cubic_spline', 'lanczos',
+                                 'average', 'mode', 'median', 'max', 'min', 'sum']
```
metasim, line 78: 🙇

metasim: Actually, do you need to do the `lower().strip('_')` thing here?

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
vpipkt: i did not see anything.

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
vpipkt: thanks for this really valuable help.

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
vpipkt: do you think we should just coerce to float if using Max as it is possible and maybe even likely to overflow??

core/src/main/scala/org/locationtech/rasterframes/extensions/RasterJoin.scala
vpipkt: ah ... in my path dependent thinking i did not even really think through having it typed in the scala api... much easier tbh

core/src/main/scala/org/locationtech/rasterframes/functions/TileFunctions.scala
vpipkt: back compatibility with the 2 argument form, yes.

core/src/main/scala/org/locationtech/rasterframes/extensions/ReprojectToLayer.scala
vpipkt: easyt

docs/src/main/paradox/reference.md
vpipkt: i am not 100% sure. not really sure what to say about this exactly. But it is certainly possible that the neighborhood that gets evaluated has a much smaller size. At least based on my reading of things.

docs/src/main/paradox/release-notes.md
vpipkt: maybe better to discuss in the raster join chapter of docs?

pyrasterframes/src/main/python/pyrasterframes/__init__.py
vpipkt: would help

docs/src/main/paradox/release-notes.md
vpipkt: take a look here

core/src/main/scala/org/locationtech/rasterframes/expressions/localops/Resample.scala
metasim: From where I sit, I think coercing to float all the time is OK, but not sure if that would irritate an analyst.

metasim:
@vpipkt As always, thanks for the fantastic work!

PR #506, opened 2020-09-14
Fixed Pythons `lazy_tiles` default value to match Scala (false).

metasim:


metasim:
Looks like Scala defaults to `true`. Misread things.

PR #521, opened 2020-11-02
Set spark master with spark.task.maxFailures in test

vpipkt:
This should reduce spurious failures on CI builds


metasim (approved):
Thanks!

PR #529, opened 2020-12-10
Improve CI docs build and dependency management

vpipkt:


pyrasterframes/build.sbt
```diff
@@ -9,7 +9,9 @@ Python / doc := (Python / doc / target).toTask.dependsOn(
   Def.sequential(
     assembly,
     Test / compile,
-    pySetup.toTask(" pweave --quick True")
+    pySetup.toTask(" develop"),
+    pySetup.toTask(" pweave --quick True"),
+    pySetup.toTask(" develop --uninstall") // put environment back how it was
```
vpipkt, line 14: It seems like it would be good  to have something like this:


```
    pySetup.toTask(" develop"),
    Try(pySetup.toTask(" pweave")),
    pySetup.toTask(" develop --uninstall")
```

So that the environment is put back how it was (more or less?) after the pweave task runs even if it fails.

That's beyond my sbt fu though

vpipkt (commented):
i'd like to get this merged before it gets real stale.

pyrasterframes/build.sbt
vpipkt: Another possibility is a much larger refactor:

1. set up tox in pyrasterframes
    1. separate environments for unit tests and docs builds
1. sbt task runs `tox  -e $TOX_ENVIRONMENT` instead of `python setup.py $SETUP_COMMAND`

This would afford a chance to also carve out many of the `setup_requires` and declare specific requirements files for the disctinct tox environments.

There is possibly some downside I am not thinking of (other than the work of refactoring), but it occurred to me that this might be a good direction

vpipkt:
Changed my mind and converted back to draft because there is still a bit of surgery to do in `setup.py` on dependencies.

PR #538, opened 2021-01-12
Use lenient=true in Z2 and XZ2 indexers

### LINKED ISSUE #531: XZ2 indexer should be lenient on coordinate values
If the tile extent is beyond (-180.0, 180.0) latitude or (-90.0, 90.0) longitude the call to [GeoMesa XZ2SFC.index](https://github.com/locationtech/geomesa/blob/master/geomesa-z3/src/main/scala/org/locationtech/geomesa/curve/XZ2SFC.scala#L42) will throw because we use [`lenient=false`](https://github.com/locationtech/rasterframes/blob/0.9.0/core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/XZ2Indexer.scala#L86).

Is there a compelling reason not to allow this index to extend beyond these bounds?

Background, originally reported here:

https://gitter.im/locationtech/geomesa?at=5fecdfc263fe0344962a53db 

https://geomesa.atlassian.net/browse/GEOMESA-2974

vpipkt:
Closes #531


duncanfwalker:
Thanks @vpipkt - that's just what I needed

Issue #501, opened 2020-08-20
rf_reproject throws on encountering null value in column


vpipkt:

netanel246:
Can I work on that?

vpipkt:
@netanel246 please have a look at the [contribution guidelines here](https://github.com/locationtech/rasterframes/blob/develop/CONTRIBUTING.md). If you have any questions about the Eclipse contributor agreement, do reach out on the [gitter chat](https://gitter.im/locationtech/rasterframes).

Feel free to reach out on gitter about your work, if you have questions or would like to discuss different implementation ideas! We are happy to have your contribution!

PR #548, opened 2021-03-10
Fix st_reproject throws on encountering null value in column

netanel246:
Author: Netanel Malka <netanel246@gmail.com>

st_reproject throws on encountering null value in the column - #501

Added null check on ReprojectGeometry
Added test for that

Signed-off-by: Netanel Malka <netanel246@gmail.com>

### DIFF
--- a/core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/ReprojectGeometry.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/ReprojectGeometry.scala
@@ -82,8 +82,13 @@ case class ReprojectGeometry(geometry: Expression, srcCRS: Expression, dstCRS: E
       // Optimized pass-through case.
       case (s: LazyCRS, r: LazyCRS) if s.encoded == r.encoded => geometry.eval(input)
       case _ =>
-        val geom = JTSTypes.GeometryTypeInstance.deserialize(geometry.eval(input))
-        JTSTypes.GeometryTypeInstance.serialize(reproject(geom, src, dst))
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/ReprojectGeometry.scala
```diff
@@ -82,8 +82,11 @@ case class ReprojectGeometry(geometry: Expression, srcCRS: Expression, dstCRS: E
       // Optimized pass-through case.
       case (s: LazyCRS, r: LazyCRS) if s.encoded == r.encoded => geometry.eval(input)
       case _ =>
-        val geom = JTSTypes.GeometryTypeInstance.deserialize(geometry.eval(input))
-        JTSTypes.GeometryTypeInstance.serialize(reproject(geom, src, dst))
+        if (geometry.eval(input) != null) {
+          val geom = JTSTypes.GeometryTypeInstance.deserialize(geometry.eval(input))
+          JTSTypes.GeometryTypeInstance.serialize(reproject(geom, src, dst))
+        }
+        null
```
metasim, line 89: Why do you always evaluate to null here... should this be in an `else`?

metasim (changes requested):

core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/ReprojectGeometry.scala
netanel246: Oops. you're right.
Fix that

metasim (approved):

PR #487, opened 2020-05-07
BufferedTile support.

metasim:
Closes #441

Issue #441, opened 2020-01-02
Implement buffered tiles


metasim:

jpolchlo:
This issue has been sitting for some time and continues to be an important milestone.  Rather than wait for the work to be done for us, we'd like to propose a work plan for getting this done.  Largely, this is based on our experience adding an experimental tensor type to a branch of RF.  A buffered version of that tensor was also provided, but it ends up being unclear if the additional dimension of the tensor type was particularly useful.  However, the experience of adding a buffered type was instructive, and we believe that introducing a buffered type will enable some valuable operations.

The main assumption behind this work is that we need a new BufferedTile type and a corresponding UDT.  The BufferedTile provided by Geotrellis is nearly worthless, so we should create a more featureful replacement.  This can be implemented directly in RF to start, and back-ported to GT if it's deemed worthy.  This base class will provide the means to work with both Tiles and BufferedTiles for use in local map algebra operations, and convolution methods should be provided for focal operations.

Expression nodes will need to be modified to allow for the two types of tiles to interact easily.  For instance, BinaryLocalRasterOp can be provided with more overloads to its op method, and the eval functions can be extended to do the work with the new type in whichever combination is required.

We'll also need to extend readers to produce buffered tiles.

Successful implementation of this feature should be accompanied by some basic focal operations.

metasim:
@jpolchlo This is fantastic. Thanks for getting the ball rolling on this. It's been on my todo for way too long.

I would propose that we seriously consider all tiles "buffered tiles", with the default buffer being zero. This would pretty much allow all operations/Expressions to continue working unchanged. However, I'm assuming a fairly naive representation of a buffered tile, where there's equal buffer amount around the whole tile, and we're only talking about adding a single field to the UDT. The GeoTrellis `BufferedTile`  uses a `GridBounds[Int]` to describe the buffering. Not sure I understand the semantics of it, and the implications of it being asymmetric.

How does this line of thinking fall apart?

Another option is to use a structural typing approach like we do for "ProjectedRasters", but I feel this is much more an implementation detail, and deserves to be hidden behind the UDT curtain.

The initial way to generate the buffered tiles would be to add a parameter to `spark.read.raster`. Then I suppose we'd want to look at `rf_reproject` and other operations where you might want to go from unbuffered to buffered whenever interpolation is involved.

Would you mind linking here your `{Buffered}TensorUDT` implementations here? Pointing out salient sections would be a bonus.

metasim:
[...]

PR #551, opened 2021-03-30
Focal Operations and Buffer Tile support

echeipesh:
This is a draft effort to add focal operations which require use of pixel buffer per tile to provide the cell neighborhood information around tile edges. This PR moves https://github.com/locationtech/rasterframes/pull/487 closer to develop and adds Python wrappers for focal operations along with elevation related operations (slope, aspect, hillshade).

Currently this PR brings over `geotrellis.raster.BufferTile` implementation for convenience of iteration. I don't think its in its final form and having it here greatly simplifies the development process. This will be moved to GeoTrellis proper at the end.

Screen Shot 2021-03-30 at 12 16 00 PM

[Notebook Demo](https://github.com/locationtech/rasterframes/blob/02b653d6e9e58f4b43dc7292e5ec7873800375ce/rf-notebook/src/main/notebooks/FocalOperations.ipynb)

## Tasks
- [x] Rebase https://github.com/locationtech/rasterframes/pull/487
- [x] Read BufferTile when requested
- [x] Add Slope, Aspect, Hillshade functions
- [x] Add python wrappers for focal functions and test in notebook 
- [x] Reconcile BufferTile with Delegate tile
- [ ] Add `Neighborhood` class and parser to pyrasterframes
- [ ] Test that tile buffer information is being correctly used by focal functions
- [ ] zFactor as Expression input (can be calculated by UDF)
- [ ] zFactor as interpolation table
- [ ] zFactor is automatic for LatLng rasters

### Delegate Tiles
The main remaining challenge is to reconcile the design of BufferTile with RasterFrame reliance of the delegate tiles. This is critical because the current pattern in GeoTrellis is to rely on matching argument tile types in order to activate special handling. For instance when combining two buffer tiles operation needs to be applied to "hidden" buffer pixels. This is not possible when both or either of the inputs are wrapped in a delegate tile. Likely pixel buffer information needs to be integrated into `Tile` interface itself.

### zFactor
For surface calculations critical input is [zFactor](https://www.esri.com/arcgis-blog/products/product/imagery/setting-the-z-factor-parameter-correctly/) which is contingent on the resolution and CRS. For a DataFrame of tiles the zFactor can change as the location of the tile shifts relative to the projection distortion. GeoTrellis has some [functionality to support this](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/mapalgebra/focal/ZFactor.scala). Some design should be lifted from [GeoPySpark API](https://github.com/locationtech-labs/geopyspark/blob/97bcb17a56ed4b4059e2f0dbab97706562cac692/geopyspark/geotrellis/hillshade.py). From user perspective there are three modes that can be helpful:

- Handle calculations automatically for supported projections (LatLng) (given elevation units)
- Use user provided interpolation table to use zFactor
- Use expression and UDF to allow user full control over zFactor given raster metadata (will require adding `rf_cellsize`)

Connects:  #441

### DIFF
--- a/bench/src/main/scala/org/locationtech/rasterframes/bench/RasterRefBench.scala
+++ b/bench/src/main/scala/org/locationtech/rasterframes/bench/RasterRefBench.scala
@@ -46,7 +46,7 @@ class RasterRefBench  extends SparkEnv with LazyLogging {
     val r2 = RFRasterSource(remoteCOGSingleband2)
 
     singleDF = Seq((r1, r2)).toDF("B1", "B2")
-      .select(RasterRefToTile(RasterSourceToRasterRefs(Some(r1.dimensions), Seq(0), $"B1", $"B2")))
+      .select(RasterRefToTile(RasterSourceToRasterRefs(Some(r1.dimensions), Seq(0), 0: Short, $"B1", $"B2")))
 
     expandedDF = Seq((r1, r2)).toDF("B1", "B2")
       .select(RasterRefToTile(RasterSourceToRasterRefs($"B1", $"B2")))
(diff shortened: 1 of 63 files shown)


metasim:
Awesome progress!

PR #554, opened 2021-06-01
Scala 2.12 and Spark 3 support

echeipesh:
This PR bumps to Spark 3 and Scala 2.12, dropping support for Scala 2.11

Two major problems faced by this PR as:
- `ExpresisonEncoder` now has expectations for what serializer and deserializer expressions look like and rewrites them on construction. Combined with getting rid of `flat` property this has introduced inconsistency in how RF encoders are treated when they're top of the column vs when they're member of a struct.
- `Filter` is now sealed class. It's no longer possible to push down spatial filter predicates to GeoTrellis relation. This is sad but probably not a huge deal in reality. Very much secondary issue.


core/src/test/scala/org/locationtech/rasterframes/encoders/DevSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+      val data = Seq(one)
+      val in: Encoder[T] = implicitly[Encoder[T]]
+      val enc = encoderFor[T](in)
+      val toRow = enc.createSerializer()
+      val attributes = enc.schema.map(f => AttributeReference(f.name, f.dataType, f.nullable, f.metadata)())
+      val encoded = data.map(d => toRow(d).copy())
+      val plan = new LocalRelation(attributes, encoded)
+      val dataset = new Dataset(spark, plan, implicitly[Encoder[T]])
+      dataset.printSchema()
+      dataset.show()
```
echeipesh, line 179: This breaks out the series of events that result 
In Spark 3 the output is:
```
root
 |-- tile_context: struct (nullable = true)
 |    |-- extent: struct (nullable = false)
 |    |    |-- xmin: double (nullable = false)
 |    |    |-- ymin: double (nullable = false)
 |    |    |-- xmax: double (nullable = false)
 |    |    |-- ymax: double (nullable = false)
 |    |-- crs: struct (nullable = false)
 |    |    |-- crsProj4: string (nullable = true)
 |-- tile: tile (nullable = false)

+--------------------+--------------------+
|        tile_context|                tile|
+--------------------+--------------------+
|{{10.0, 20.0, 30....|ArrayTile(10,10,i...|
```

In spark 2:

```
root
 |-- value: struct (nullable = true)
 |    |-- tile_context: struct (nullable = true)
 |    |    |-- extent: struct (nullable = false)
 |    |    |    |-- xmin: double (nullable = false)
 |    |    |    |-- ymin: double (nullable = false)
 |    |    |    |-- xmax: double (nullable = false)
 |    |    |    |-- ymax: double (nullable = false)
 |    |    |-- crs: struct (nullable = false)
 |    |    |    |-- crsProj4: string (nullable = false)
 |    |-- tile: tile (nullable = false)

+--------------------+
|               value|
+--------------------+
|[[[10.0, 20.0, 30...|
|[[[10.0, 20.0, 30...|
+--------------------+
```

core/src/test/scala/org/locationtech/rasterframes/encoders/DevSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+      val data = Seq((one, one))
+      val in: Encoder[T] = implicitly[Encoder[T]]
+      val enc = encoderFor[T](in)
+      val toRow = enc.createSerializer()
+      val attributes = enc.schema.map(f => AttributeReference(f.name, f.dataType, f.nullable, f.metadata)())
+      val encoded = data.map(d => toRow(d).copy())
+      val plan = new LocalRelation(attributes, encoded)
+      val dataset = new Dataset(spark, plan, implicitly[Encoder[T]])
+      dataset.printSchema()
+      dataset.show()
```
echeipesh, line 197: In Spark 3:
```
root
 |-- _1: struct (nullable = false)
 |    |-- tile_context: struct (nullable = true)
 |    |    |-- extent: struct (nullable = false)
 |    |    |    |-- xmin: double (nullable = false)
 |    |    |    |-- ymin: double (nullable = false)
 |    |    |    |-- xmax: double (nullable = false)
 |    |    |    |-- ymax: double (nullable = false)
 |    |    |-- crs: struct (nullable = false)
 |    |    |    |-- crsProj4: string (nullable = true)
 |    |-- tile: tile (nullable = true)
 |-- _2: struct (nullable = false)
 |    |-- tile_context: struct (nullable = true)
 |    |    |-- extent: struct (nullable = false)
 |    |    |    |-- xmin: double (nullable = false)
 |    |    |    |-- ymin: double (nullable = false)
 |    |    |    |-- xmax: double (nullable = false)
 |    |    |    |-- ymax: double (nullable = false)
 |    |    |-- crs: struct (nullable = false)
 |    |    |    |-- crsProj4: string (nullable = true)
 |    |-- tile: tile (nullable = true)

+--------------------+--------------------+
|                  _1|                  _2|
+--------------------+--------------------+
|{{{10.0, 20.0, 30...|{{{10.0, 20.0, 30...|
+--------------------+--------------------+
```

in Spark 2:
```
root
 |-- _1: struct (nullable = true)
 |    |-- tile_context: struct (nullable = true)
 |    |    |-- extent: struct (nullable = false)
 |    |    |    |-- xmin: double (nullable = false)
 |    |    |    |-- ymin: double (nullable = false)
 |    |    |    |-- xmax: double (nullable = false)
 |    |    |    |-- ymax: double (nullable = false)
 |    |    |-- crs: struct (nullable = false)
 |    |    |    |-- crsProj4: string (nullable = false)
 |    |-- tile: tile (nullable = false)
 |-- _2: struct (nullable = true)
 |    |-- tile_context: struct (nullable = true)
 |    |    |-- extent: struct (nullable = false)
 |    |    |    |-- xmin: double (nullable = false)
 |    |    |    |-- ymin: double (nullable = false)
 |    |    |    |-- xmax: double (nullable = false)
 |    |    |    |-- ymax: double (nullable = false)
 |    |    |-- crs: struct (nullable = false)
 |    |    |    |-- crsProj4: string (nullable = false)
 |    |-- tile: tile (nullable = false)

+--------------------+--------------------+
|                  _1|                  _2|
+--------------------+--------------------+
|[[[10.0, 20.0, 30...|[[[10.0, 20.0, 30...|
+--------------------+--------------------+
```

core/src/test/scala/org/locationtech/rasterframes/encoders/DevSpec.scala
echeipesh: This behavior breaks `.as[ProjectedRasterTile]` both when used on the output of DataFrame and when used on the column type conversion from DataFrame to DataSet. It seems to be the root cause of many/most/all failed tests where assignment to `Tuple1` during expression evaluation fails because ProjectedRasterTile gets blown out in two columns instead of one.

core/src/test/scala/org/locationtech/rasterframes/encoders/DevSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+      val ds = Seq(Person("Bob", 11), Person("Eugene", 10)).toDS
+
+      ds.printSchema()
+      ds.show()
+
+      val df = ds.toDF()
+
+      val me = df.as[Person].first()
+      info(me.toString)
+    }
```
echeipesh, line 69: This is an attempt to mirror and break up the logic of construction ExpressionEncoder from spark `ScalaReflection` to see if we can build a type-class driven way to compose the encoders similar to CatalystSerializer. In this case the encoder for `Person` and `School` where built up "by hand" in this manner showing that it's at least possible.

The key motivator for this exercise is that the way spark constructs serializer/deserializer for case classes behaves well with regard to the problem we're seeing with ProjectedRasterTile. That is, it handles blowing out the members of Person struct to columns when its at row level and then recombining them into the java type for DataSet.

core/src/test/scala/org/locationtech/rasterframes/encoders/DevSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+      }
+      val ds = Seq(Person("Bob", 11), Person("Eugene", 10)).toDS
+
+      ds.printSchema()
+      ds.show()
+
+      val df = ds.toDF()
+
+      val me = df.as[Person].first()
+      info(me.toString)
```
echeipesh, line 68:
```
Person(Bob,11)

root
 |-- name: string (nullable = true)
 |-- age: integer (nullable = true)

+------+---+
|  name|age|
+------+---+
|   Bob| 11|
|Eugene| 10|
+------+---+
```

core/src/test/scala/org/locationtech/rasterframes/encoders/DevSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+
+      ds.printSchema()
+      ds.show()
+
+      val df = ds.toDF()
+
+      val it = df.as[School].first()
+      info(it.toString)
+    }
+  }
```
echeipesh, line 89:
```
School(Person(Sid,0),Person(Eugene,10))

root
 |-- teacher: struct (nullable = true)
 |    |-- name: string (nullable = true)
 |    |-- age: integer (nullable = true)
 |-- student: struct (nullable = true)
 |    |-- name: string (nullable = true)
 |    |-- age: integer (nullable = true)

+--------+------------+
| teacher|     student|
+--------+------------+
|{Sid, 0}|{Eugene, 10}|
+--------+------------+
```

core/src/test/scala/org/locationtech/rasterframes/encoders/DevSpec.scala
echeipesh: Little more context here: It seems that "value" nesting is now reserved for things that aren't `StructType` (primitives and arrays) https://github.com/apache/spark/blob/1d550c4e90275ab418b9161925049239227f3dc9/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/encoders/ExpressionEncoder.scala#L262-L267

Quick experiment of wrapping the ProjectedRasterTile in `value` struct on our end yielded predictable results of fixing the "one" test and breaking the "two" test but preserving the intermediate `value` nesting under each column.

core/src/test/scala/org/locationtech/rasterframes/functions/LocalFunctionsSpec.scala
```diff
@@ -38,10 +38,10 @@ class LocalFunctionsSpec extends TestEnvironment with RasterMatchers {
     it("should local_add") {
       val df = Seq((one, two)).toDF("one", "two")
 
-      val maybeThree = df.select(rf_local_add($"one", $"two")).as[ProjectedRasterTile]
-      assertEqual(maybeThree.first(), three)
+      val maybeThree = df.select(rf_local_add($"one", $"two")).as[Option[ProjectedRasterTile]]
+      assertEqual(maybeThree.first().get, three)
```
metasim, line 42: I'm totally fine with this, but I'm afraid if @pomadchin ever sees this he'll call in the functional purity police 🚔 😆

echeipesh: `.get` for the test only, no harm!

pomadchin: @metasim @echeipesh for _tests_ only 👁️

metasim: @pomadchin where's your work with using Frameless for encoding?

pomadchin: @metasim diff is here https://github.com/echeipesh/rasterframes/pull/2/files, I think @echeipesh continues the work using my branch. But it's not pushed yet cc @echeipesh

metasim: Awesome... Any chance the work could be coalesced into this PR? Nice to be able to see the progress and have a common view of where the current blockers are.

metasim: Or is the issue here that that STAC DataSource is interleaved with this work?

pomadchin:
@echeipesh @metasim How about merging it as is without fixing the Python for now to unblock #551?

metasim:
> @echeipesh @metasim How about merging it as is without fixing the Python for now to unblock #551?

Let's go for it! I'll work the python fixes on another PR.

metasim:
@echeipesh @pomadchin Congratulations!! 🎉 Amazing work!!

pomadchin:
Closes #530 as well!

PR #562, opened 2021-09-17
Python binding updates for PySpark 3.1

metasim:


pyrasterframes/src/main/python/pyrasterframes/version.py
```diff
@@ -20,4 +20,4 @@
 #
 
 # Translating Java version from version.sbt to PEP440 norms
-__version__: str = '0.9.1'
+__version__: str = '0.9.2'
```
pomadchin, line 23: I vote for 0.10.0! 🚀

metasim: I think that's smart. Let's keep that in mind. I tried to keep a single point of authority with the version number, but it kept drifting away with Python :(

PR #567, opened 2021-10-01
Add Focal Operations TargetCell support

pomadchin:
This PR allows to specify the focal operations target cells: data, nodata, all.

TODO:
- [x] Modify the Jupyter Notebook demo
- [x] Docker image for tests: quay.io/daunnc/rasterframes-notebook:0.9.2-SNAPSHOT

image


core/src/main/scala/org/apache/spark/sql/rf/QuinaryExpression.scala
```diff
@@ -0,0 +1,111 @@
...
+
+import org.apache.spark.sql.catalyst.expressions.codegen.Block._
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.catalyst.expressions.codegen.{CodeGenerator, CodegenContext, ExprCode, FalseLiteral}
+
+/**
+ * An expression with five inputs and one output. The output is by default evaluated to null if any input is evaluated to null
+ */
+abstract class QuinaryExpression extends Expression {
```
pomadchin, line 11: There is only `Quaternary` and `Septenary` in the spark core.

metasim: I think those might be new! I think it topped out at trinary before?

core/src/main/scala/org/apache/spark/sql/rf/QuinaryExpression.scala
```diff
@@ -0,0 +1,111 @@
...
+    if (value1 != null) {
+      val value2 = exprs(1).eval(input)
+      if (value2 != null) {
+        val value3 = exprs(2).eval(input)
+        if (value3 != null) {
+          val value4 = exprs(3).eval(input)
+          if (value4 != null) {
+            val value5 = exprs(4).eval(input)
+            if (value5 != null) {
+              return nullSafeEval(value1, value2, value3, value4, value5)
```
metasim, line 33: 😱 😆 
Surprised there's not a monad for that!

core/src/main/scala/org/apache/spark/sql/rf/QuinaryExpression.scala
pomadchin: @metasim I think you're 💯 correct

metasim (approved):
Nice!

PR #569, opened 2021-10-04
Updates Python documentation build to work with Spark 3.x, etc.

metasim:


pyrasterframes/src/main/python/pyrasterframes/__init__.py
```diff
@@ -255,19 +255,26 @@ def temp_name():
 def _stac_api_reader(
         df_reader: DataFrameReader,
         uri: str,
+        flatten_assets: bool = True,
```
pomadchin, line 258: how about making it the last argument of a function? Seems to me a bit more obvious: `uri, filter, flattening flag`

PR #573, opened 2021-11-09
Fix UDTs registration ordering

pomadchin:
Without this fix workers may try to access Encoders before registering UDTs.

Tested both 66cca65 and 8b5c165 within the k8s notebook.

The image to play with is ` quay.io/daunnc/rasterframes-notebook:0.10.1-SNAPSHOT`

### DIFF
--- a/core/src/main/scala/org/locationtech/rasterframes/ref/RFRasterSource.scala
+++ b/core/src/main/scala/org/locationtech/rasterframes/ref/RFRasterSource.scala
@@ -31,7 +31,7 @@ import geotrellis.vector.Extent
 import org.apache.hadoop.conf.Configuration
 import org.apache.spark.annotation.Experimental
 import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder
-import org.apache.spark.sql.rf.{CrsUDT, RasterSourceUDT, TileUDT}
+import org.locationtech.rasterframes.encoders.StandardEncoders
 import org.locationtech.rasterframes.model.TileContext
 import org.locationtech.rasterframes.{NOMINAL_TILE_DIMS, rfConfig}
 
@@ -100,13 +100,7 @@ object RFRasterSource extends LazyLogging {
 
   def cacheStats = rsCache.stats()
 
-  implicit def rsEncoder: ExpressionEncoder[RFRasterSource] = {
-    // Makes sure UDTs are registered first
-    RasterSourceUDT
-    TileUDT
-    CrsUDT
-    ExpressionEncoder()
-  }
+  implicit lazy val rsEncoder: ExpressionEncoder[RFRasterSource] = StandardEncoders.rfRasterSourceEncoder
 
   def apply(source: URI): RFRasterSource =
     rsCache.get(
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/locationtech/rasterframes/ref/RFRasterSource.scala
```diff
@@ -101,7 +101,10 @@ object RFRasterSource extends LazyLogging {
   def cacheStats = rsCache.stats()
 
   implicit def rsEncoder: ExpressionEncoder[RFRasterSource] = {
-    RasterSourceUDT // Makes sure UDT is registered first
+    // Makes sure UDTs are registered first
+    RasterSourceUDT
+    TileUDT
+    CrsUDT
```
pomadchin, line 107: Playing safe here and registering everything.

core/src/main/scala/org/locationtech/rasterframes/tiles/ProjectedRasterTile.scala
```diff
@@ -58,5 +59,11 @@ object ProjectedRasterTile {
 
   def unapply(prt: ProjectedRasterTile): Option[(Tile, Extent, CRS)] = Some((prt.tile, prt.extent, prt.crs))
 
-  implicit lazy val projectedRasterTileEncoder: ExpressionEncoder[ProjectedRasterTile] = ExpressionEncoder()
+  implicit lazy val projectedRasterTileEncoder: ExpressionEncoder[ProjectedRasterTile] = {
+    // Makes sure UDTs are registered first
+    RasterSourceUDT
+    TileUDT
+    CrsUDT
```
pomadchin, line 66: Playing safe here and registering everything.

pyrasterframes/src/main/python/setup.py
```diff
@@ -140,7 +140,7 @@ def dest_file(self, src_file):
 # to throw a `NotImplementedError: Can't perform this operation for unregistered loader type`
 pytest = 'pytest>=4.0.0,<5.0.0'
 
-pyspark = 'pyspark==3.1.1'
+pyspark = 'pyspark==3.1.2'
```
pomadchin, line 143: Don't know how it worked before 🤦

rf-notebook/src/main/docker/requirements-nb.txt
```diff
@@ -1,4 +1,4 @@
-pyspark>=3.1
+pyspark==3.1.2
```
pomadchin, line 1: Spark 3.2 has been released and we need more strict constraints here.

echeipesh:
Screen Shot 2021-11-10 at 8 40 43 AM

Whats up with the CI failure here? Have you seen something like this? (Hey Google)

metasim:
@pomadchin Do you know why the difference between `3.1.1` and `3.1.2` made a difference?

metasim:
@pomadchin @echeipesh What do you think about sticking with CircleCI until we have time to figure out GitHub actions?

metasim (approved):
Thanks @pomadchin !

pomadchin:
@metasim Nope don't know, I noticed that even minor spark version mismatch on a driver and workers may cause to fail even jobs scheduling. I dont know why it worked before 🦐 but worked: I guess it matters what spark version is installed on a host and it 'just works'.

metasim:
> I dont know why it worked before 🦐 but worked: I guess it matters what spark version is installed on a host and it 'just works'.

Maybe only triggered in cluster mode ¯\_(ツ)_/¯

core/src/main/scala/org/locationtech/rasterframes/tiles/ProjectedRasterTile.scala
metasim: Can this be done in a less ad-hoc manner? At least have a single function across the source that implements this pattern?

PR #582, opened 2022-04-07
CI fix.

metasim:
Addresses [this](https://app.circleci.com/pipelines/github/locationtech/rasterframes/870/workflows/df3eced6-5b89-45af-9a19-810aaaaf6f00/jobs/2589).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,7 @@
  *
  */
 
-// Leave me an my custom keys alone!
+// Leave me and my custom keys alone!
 Global / lintUnusedKeysOnLoad := false
 
 addCommandAlias("makeSite", "docs/makeSite")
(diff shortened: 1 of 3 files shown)


metasim:
@pomadchin it seems CircleCI integration has disappeared. Have you seen this elsewhere? Not sure what happened.

pomadchin:
@metasim yea, so I think it is a change in the default checks settings for all github projects. The way to fix it is to go to cricle `Organization Settings` => `VSC` => `Manage GitHub Checks`

i.e. 
image

pomadchin:
That sounds like we need to file an ipzilla ticket ._.

pomadchin (commented):
Could we also bump the GT version up and STAC4s (it contains a serious [pagination related improvement](https://github.com/azavea/stac4s/blob/master/CHANGELOG.md#080---2022-03-30))?

It can also be a separate PR though.

PR #587, opened 2022-06-30
Spark 3.2

echeipesh:
Pushing for Spark 3.2
- [x] get tests green
- [x] update to spark-testing-base for more speed
- [x] get code review and/or cleanup the changes
- [x] figure out what to do with GeoTrellis layer data source
- [x] update python package
  - ~cleanup dependencies: disentangle the release wheel from pweave~
  - ~grok the python environment, consider introducing `poetry` instead of conda~
- [x] run some jobs in spark 3.2 environment


core/src/main/scala/org/locationtech/rasterframes/expressions/BinaryRasterFunction.scala
```diff
@@ -25,13 +25,14 @@ import com.typesafe.scalalogging.Logger
...
-import org.apache.spark.sql.catalyst.expressions.BinaryExpression
+import org.apache.spark.sql.catalyst.expressions.{BinaryExpression, Expression}
 import org.apache.spark.sql.types.DataType
 import org.locationtech.rasterframes.expressions.DynamicExtractors._
 import org.slf4j.LoggerFactory
 
 /** Operation combining two tiles or a tile and a scalar into a new tile. */
-trait BinaryRasterFunction extends BinaryExpression with RasterResult {
+trait BinaryRasterFunction extends BinaryExpression with RasterResult { self: HasBinaryExpressionCopy =>
+  override protected def withNewChildrenInternal(newLeft: Expression, newRight: Expression): Expression = copy(newLeft, newRight)
```
echeipesh: Tried being clever here with structural self type in order to call copy in the super-type and I really don't think this is worth the brain cells. Down this road is creating things like `RFBinaryExpression` that has this little bit of magic. However, it will not always work, for instance when the case class does not use standard `left`, `right` or `child` constructor names. So I'm probably going to revert this and do some more c/p.

core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/ReprojectGeometry.scala
```diff
@@ -49,12 +49,12 @@ import org.locationtech.rasterframes.model.LazyCRS
     > SELECT _FUNC_(geom, srcCRS, dstCRS);
        ..."""
 )
-case class ReprojectGeometry(geometry: Expression, srcCRS: Expression, dstCRS: Expression) extends Expression with CodegenFallback {
-
+case class ReprojectGeometry(geometry: Expression, srcCRS: Expression, dstCRS: Expression) extends TernaryExpression with CodegenFallback {
   override def nodeName: String = "st_reproject"
+  def first: Expression = geometry
+  def second: Expression = srcCRS
+  def third: Expression = dstCRS
```
echeipesh, line 56: Couple of places the Expression case classes actually have meaningful names on the constructor args which feels nice. Where they don't the naming convention changed.

pomadchin:
I'm going to bump the min Spark version in GT as well https://github.com/locationtech/geotrellis/pull/3389 
Thogh it should not affect this PR much.

echeipesh:
@pomadchin It'd be nice to get 3.6.3 out with spark 3.2.0 and then 3.7.0 with CES3 and spark 3.2.1. That'd allow this to sync with frameless 0.11 and later with 0.12+.

pomadchin:
@echeipesh 🚀 https://github.com/locationtech/geotrellis/pull/3471 + I'm dropping Spark 3.0 over there https://github.com/typelevel/frameless/pull/637 (frameless will be published for 3.1.x, 3.2.x, and 3.3.x)

datasource/src/test/scala/org/locationtech/rasterframes/datasource/geotrellis/GeoTrellisDataSourceSpec.scala
```diff
@@ -48,7 +48,8 @@ import org.scalatest.{BeforeAndAfterAll, Inspectors}
 
 import scala.math.{max, min}
 
-class GeoTrellisDataSourceSpec extends TestEnvironment with BeforeAndAfterAll with Inspectors with DataSourceOptions {
+trait GeoTrellisDataSourceSpec extends TestEnvironment with BeforeAndAfterAll with Inspectors with DataSourceOptions {
```
echeipesh, line 51: Very little motivation to fix this. Note that the push-down predicates also got busted with move to spark 3.2. 

Can't imagine ever using this rather than reading COGs or even Zarr. Unless somebody would like to fix this I would like to suggest to remove this feature altogether to keep things tidy.

echeipesh: @metasim @pomadchin What do you guys think?

project/RFDependenciesPlugin.scala
```diff
@@ -70,9 +71,9 @@ object RFDependenciesPlugin extends AutoPlugin {
...
     ),
     // NB: Make sure to update the Spark version in pyrasterframes/python/setup.py
-    rfSparkVersion := "3.1.3",
-    rfGeoTrellisVersion := "3.6.1",
-    rfGeoMesaVersion := "3.2.0",
-    excludeDependencies += "log4j" % "log4j"
+    rfSparkVersion := "3.2.1",
+    rfGeoTrellisVersion := "3.6.3",
+    rfGeoMesaVersion := "3.4.1"
+    //excludeDependencies += "log4j" % "log4j"
```
echeipesh: Wait, why was this excluded?

pomadchin: I think it's another way to solve the CVE problem but idk!

core/src/main/scala/org/locationtech/rasterframes/expressions/generators/RasterSourceToTiles.scala
```diff
@@ -84,6 +84,8 @@ case class RasterSourceToTiles(children: Seq[Expression], bandIndexes: Seq[Int],
         Traversable.empty
     }
   }
+
+  override protected def withNewChildrenInternal(newChildren: IndexedSeq[Expression]): Expression = copy(children=newChildren)
```
pomadchin: minor: seems the formatting is a bit off there `children=newChildren` => `children = newChildren`

datasource/src/test/scala/org/locationtech/rasterframes/datasource/geotrellis/GeoTrellisDataSourceSpec.scala
pomadchin: :+1:

core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
```diff
@@ -60,114 +64,166 @@ package object expressions {
...
+
+    def register3[T <: Expression : ClassTag](
+      name: String,
+      builder: (Expression, Expression, Expression) => T
+    ): Unit = registerFunction[T](name, None){ args => builder(args(0), args(1), args(2)) }
+
+    def register5[T <: Expression : ClassTag](
+      name: String,
+      builder: (Expression, Expression, Expression, Expression, Expression) => T
+    ): Unit = registerFunction[T](name, None){ args => builder(args(0), args(1), args(2), args(3), args(4)) }
```
pomadchin: So far the only ugly place!

pomadchin (approved):
LGTM!

core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
```diff
@@ -46,6 +46,10 @@ import scala.reflect.runtime.universe._
  * @since 10/10/17
  */
 package object expressions {
+  type HasTernaryExpressionCopy = {def copy(first: Expression, second: Expression, third: Expression): Expression}
+  type HasBinaryExpressionCopy = {def copy(left: Expression, right: Expression): Expression}
+  type HasUnaryExpressionCopy = {def copy(child: Expression): Expression}
```
pomadchin: Those can be expensive I think but mb it's fine 🤔

pomadchin:
CI is dead? https://app.circleci.com/pipelines/github/locationtech/rasterframes?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

core/src/main/scala/org/locationtech/rasterframes/expressions/generators/RasterSourceToTiles.scala
echeipesh: Heh, python formatting bleeding through.

metasim (approved):

core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/MaskExpression.scala
```diff
@@ -0,0 +1,74 @@
...
+import geotrellis.raster.Tile
+import org.apache.spark.sql.catalyst.analysis.TypeCheckResult
+import org.apache.spark.sql.catalyst.analysis.TypeCheckResult.{TypeCheckFailure, TypeCheckSuccess}
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.types.DataType
+import org.locationtech.rasterframes.expressions.DynamicExtractors.tileExtractor
+
+import spire.syntax.cfor._
+
+trait MaskExpression { self: Expression =>
```
pomadchin, line 33: We can cut a fresh GT release with the fixed mask operation! Sadly due to the CircleCI breach the most up to date snapshot has not been released.

pomadchin (approved):

core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
echeipesh: What do you mean by expensive here?

core/src/main/scala/org/locationtech/rasterframes/expressions/transformers/MaskExpression.scala
echeipesh: Actually this is kind of fine. The logic here is slightly different from the standard implementation because we know we're mapping over int values. I would bet that it would be more helpful in the future to have full control over this than saving a few lines of code.

core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
echeipesh: Thank you for fixing it!

core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
pomadchin: Structural types in Scala2 rely on runtime reflection; that's not a big deal but it just got caught by my eye.

pomadchin: A bit better practice is usually to avoid them by having named traits 💭

core/src/main/scala/org/locationtech/rasterframes/expressions/BinaryRasterFunction.scala
echeipesh: Yes, this was a bad idea

core/src/main/scala/org/locationtech/rasterframes/expressions/package.scala
echeipesh: Thank you for calling this out. Can't have named traits because it's trying to call a copy method that will be added by the case class. So it HAS to be runtime reflection. But the whole point of this API change was that runtime reflection was killing Catalyst performance. So while its ugly its better to just go ahead and implement the `withNewChildrenInternal` methods in the case class definitions.

echeipesh:
I've been looking at the python side of the project enough to form strong enough opinions of it. I think it's actually going to need some work and it doesn't make sense to do it in this PR.

- make the build PEP 517 compatible
- ditch use of `conda` for `poetry` (we're not building a conda package here anyway)
- separate the the `pweave` deps from the release deps
- decide if shipping jars with `.whl` still makes sense
  - it was a nice attempt at nicety but unless you start the process from `ipython` instead of `pyspark` it is just dead weight.
- `pyrasterframes` python files are in the sbt project structure
  - Kind of the only benefit of this is to couple the production of assembly jar with the building of the .whl
  - Not sure the weirdness of this justifies the setup

So that's a lot of moving stuff around potentially and it would be much cleaner to make that it's own PR.

PR #600, opened 2023-02-25
Improve Python build process and streamline CI

thomas-maschler:
Changes made:
- update to pyspark 3.3.2
- pull out python code into `/python` folder
- use `poetry` and `pyproject.toml` to manage dependencies and project metadata
- use pytest for testing
- decouple build documentation from build wheel process
- use `Makefile` to manage main dev and CI events for both Scala and Python
- single out Scala Build, Python Build and Artifact upload into separate CI jobs
- Lint Python code using `Black` and `iSort`

Note: This PR allows to build python wheel separately from the python documentation. Building the Python Markdown documentation still fails due to invalid URLs in the docs. Fixing the Python documentation should be covered in its own PR.


python/tests/GeotrellisTests.py
```diff
@@ -0,0 +1,75 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+import os
+import pathlib
+import shutil
+import tempfile
+
+import pytest
+
+on_circle_ci = os.environ.get("CIRCLECI", "false") == "true"
```
thomas-maschler: Is CircleCI still in use?

python/tests/conftest.py
```diff
@@ -0,0 +1,131 @@
...
+            except OSError:
+                pass
+    except ImportError:
+        pass
+
+
+_chmodit()
+
+jar_dir = Path(".") / "dist"
+jar_path = next(jar_dir.glob("*assembly*.jar"))
```
thomas-maschler, line 51: I removed all the logic to dynamically look up the Assembly Jar and expect it to be present in the JAR Path or explicitly set via spark config.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,122 @@
...
+#      - uses: actions/cache@v3
+#        with:
+#          path: ./dist/*
+#          key: dist-${{ github.sha }}
+
+# TODO: Where does this go?
+#      - name: upload wheel
+#        working-directory: dist
+#        shell: bash
+#        run:
```
thomas-maschler, line 122: @echeipesh what is the correct workflow here?

pyrasterframes/src/main/python/pyrasterframes/utils.py
```diff
@@ -1,102 +0,0 @@
...
-
-def find_pyrasterframes_assembly() -> Union[bytes, str]:
-    jar_dir = find_pyrasterframes_jar_dir()
-    jarpath = glob.glob(os.path.join(jar_dir, 'pyrasterframes-assembly*.jar'))
-
-    if not len(jarpath) == 1:
-        raise RuntimeError(f""" 
-Expected to find exactly one assembly in '{jar_dir}'. 
-Found '{jarpath}' instead.""")
-    return jarpath[0]
```
thomas-maschler, line 58: I deleted these methods in the new location.

python/tests/GeotrellisTests.py
echeipesh: nope

.github/workflows/ci.yml
echeipesh: The `actions/cache@v3` is an optimization to pull the assembly from workflow cache. We don't need it. It's alright or even better to rebuild it.

I'm not sure where the wheel goes yet, PyPi eventually but I'll need to figure out how that happens.

the artefacts bit is for debugging failed CI. I never used that part of the CI. So much has changed and I think will change a little more that I'm not worried about it. I'm sure when we need it we will wire it right up.

Overall I think the CI is good and even green. I'm a little surprised we didn't end up needing GDAL for the tests though.

echeipesh (approved):
❤️

PR #602, opened 2023-03-08
match dependencies with geotrellis 3.6.3

echeipesh:


echeipesh:
Need to iron away `sttp`:

```
[info] compiling 38 Scala sources to /Users/eugene/proj/rasterframes/core/target/scala-2.12/test-classes ...
[error] stack trace is suppressed; run last datasource / update for the full output
[error] (datasource / update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error]
[error] 	* org.scala-lang.modules:scala-java8-compat_2.12:1.0.0 (early-semver) is selected over 0.8.0
[error] 	    +- com.github.blemale:scaffeine_2.12:4.1.0            (depends on 1.0.0)
[error] 	    +- com.softwaremill.sttp.client3:core_2.12:3.6.2      (depends on 0.8.0)
```

